Issue #188, opened 2011-08-25
JQUERY Validation Plugin bug -- rule of min(value) does not respect the minimum value, if the value contains commas 


rchakravarti:
Reference:  http://docs.jquery.com/Plugins/Validation/Methods/min#value
Validation Plugin Version 1.8.1
jQuery Core Version 1.5.1

To reproduce the bug:
On the Demo for the Validate a minimum of 13,
1) Enter the value 1,300 (1300 with a comma) with
the "min" rule in place, validation fails

Any number with a comma fails validation when the
"min" rule is in place, the validate plugin incorrectly reports an error

2) Remove the "min" requirement from the rule, and enter
1,300 (with the comma) -- the error goes away

Expected behavior:
The plugin should treat numbers with commas in  the same 
manner as numbers without commas.

---

The same issue was encountered using the "range" rule

Appreciate any reply to help on this issue

piersh:
this is a joke. both issues #188 and #379 which have been unfixed for years are closed referring to each other as reasons for closing. yet this fix STILL has not made it into release.

joeRinehart:
In case anyone hits this and needs a quick, U.S.-but-customizable fix, I already had numeral.js involved and using it to preformat values was a pinch:

https://gist.github.com/joeRinehart/0e8af523fdfc567cc670ddca2eaff137

PR #383, opened 2012-04-19
Added comma replacement for min/max methods and updated tests. Fixes #379

### LINKED ISSUE #379: Thousands separators allowed for numbers, but not min/max
While trying to validate a number field with min and max attributes, I noticed that there was no invalid number complaint for something like "9,000" but that adding commas would cause the value to incorrectly fail min/max validation.

mlynch:
It makes sense that if we accept numbers with US comma separators for numbers that we also accept them for min/max.


piersh:
this is a joke. both issues #188 and #379 which have been unfixed for years are closed referring to each other as reasons for closing. yet this fix STILL has not made it into release.

PR #637, opened 2013-02-12
Search for an item to compare in context of form

sievlev:


staabm:
Is there still interest in getting this PR landed?

Issue #401, opened 2012-04-25
Javascript error at the staticRules method while validating form in IE


ruado1987:
I have encountered a JavaScript error because this very line: var validator = $.data(element.form, 'validator');  in the staticRules method. Somehow it sometimes returns null, thereby resulting in error when executing the next line. This problem only occurs in IE, including IE 6,7,8. I am using jquery validation 1.9 with jquery 1.7.1

afilina:
@prgwiz Thanks for the tip. This actually didn't work for me for some reason, so I used this variation instead:
```
staticRules: function( element ) {
    var rules = {};
    var validator = $.data(element.form, "validator");
    if ( validator && validator.settings.rules ) {
      rules = $.validator.normalizeRule(validator.settings.rules[element.name]) || {};
    }
    return rules;
},
```
If anyone can tell me why validator is sometimes null despite having element.form, that will be great.

staabm:
@afilina could you provide a simplified jsfiddle which demonstrates a situation in which it doesnt work?

afilina:
@staabm Well someone is deleting my comments, so no, I won't spend hours putting up a jsfiddle. I'll just hack the core myself and use it in my legacy projects.

staabm:
> Well someone is deleting my comments

hmm? I can believe that someone intentionally deletes comments in this tracker..

> I'll just hack the core myself and use it in my legacy projects.

feel free to PR your changes, so we have can discuss something.

samuelsethhildebrand:
staticRules: function( element ) {
		var rules = {};
		
		if (!element.form)
		{ validator = $(element).parents('form').data('validator');
		}
		else
		{ validator = $.data( element.form, "validator" );
		}
		
		if ( validator.settings.rules ) {
			rules = $.validator.normalizeRule( validator.settings.rules[ element.name ] ) || {};
		}
		
		return rules;
	},

samuelsethhildebrand:
I updated my post to include more background on my changes.

PR #704, opened 2013-03-23
Fixes issue 401 which causes error when nested form created programatica...

ruado1987:
@jzaefferer: This PR replaces issue #401.

### DIFF
--- a/jquery.validate.js
+++ b/jquery.validate.js
@@ -480,6 +480,11 @@ $.extend($.validator, {
 					console.error( "%o has no name assigned", this);
 				}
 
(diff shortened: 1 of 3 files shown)


benqus:
Indeed, it is invalid in HTML5 but the problem is that other libraries generate a form  into the DOM dynamically, such as plupload (https://github.com/moxiecode/plupload/blob/1.5.7/src/javascript/plupload.html4.js#L89).

Our case is an `<input type="file" />` which we - frankly - don't care about validating.

Maybe just ignore validations of fields from nested forms?

PR #710, opened 2013-03-27
Fixed error thrown when element missing name

nberardi:
I have the following code

``` html
<form id="myform">
    <input type="text" name="name" placeholder="Full Name" required />
    <label class="checkbox">
        <input type="checkbox" required />
        I agree not to send any unsolicited email, or send email to addresses purchased, rented or otherwise obtained from a third party. I understand that port 25 is blocked by default and I have to request that it be opened. I also understand that if I violate the rules set by Server Blaze that port 25 will be closed.
    </label>
    <button type="submit">Done</button>
</form>
```

and in JavaScript I have this

``` javascript
$("#myform").validate()
```

The code always errors out and continues with submit regardless of if a name is entered into the name field or not.  I have tracked the error down to this line:

https://github.com/jzaefferer/jquery-validation/blob/master/jquery.validate.js#L729

With this PR, I fixed the error in the code by skipping the name check if the name doesn't exit.

### DIFF
--- a/jquery.validate.js
+++ b/jquery.validate.js
@@ -725,7 +725,7 @@ $.extend($.validator, {
 
 		validationTargetFor: function( element ) {
 			// if radio/checkbox, validate first element in group instead
-			if ( this.checkable(element) && element.name != "" ) {
+			if ( this.checkable(element) && element.name !== "" ) {
 				element = this.findByName( element.name ).not(this.settings.ignore)[0];
 			}
 			return element;
--- a/jquery.validate.js
+++ b/jquery.validate.js
@@ -725,7 +725,7 @@ $.extend($.validator, {
 
 		validationTargetFor: function( element ) {
 			// if radio/checkbox, validate first element in group instead
-			if ( this.checkable(element) ) {
+			if ( this.checkable(element) && element.name != "" ) {
 				element = this.findByName( element.name ).not(this.settings.ignore)[0];
 			}
 			return element;


jzaefferer:
Looking at this again, the patch doesn't make much sense.

nberardi:
@jzaefferer why doesn't the patch make sense? If the name is empty then you cannot do a findByName.

jzaefferer:
I may have misread then. Will check it out again, later.

staabm:
Applying this fix as is would also mean that `ignore` rules wont apply for those elements.

It feels like the name attribute should be required instead.

jzaefferer:
Well, this plugin requires it anyway, for anything with a rule attached. In this case, the non-named input has a required attribute, so the plugin requires a name attribute.

I've started labelling related issues: https://github.com/jzaefferer/jquery-validation/labels/name-attribute

staabm:
> Well, this plugin requires it anyway, for anything with a rule attached. In this case, the non-named input has a required attribute, so the plugin requires a name attribute.

Right but the PR is targeted to change that for checkboxes/radios. I think we agree that name will stay  mandatory and therefore this PR is invalid.

nberardi:
@staabm @jzaefferer if the name is required, which I am not arguing with that decision.  Things should fail more gracefully then they do right now.  Because if somebody is using jquery-validation ontop of a straight HTML5 implementation with no custom configuration.  The code just bombs out and the developer is left wondering why his code isn't working.  

If the name attribute is truly required, which is not HTML5 purist friendly, then some kind of error should be thrown into the console to alert the developer as to why his code is failing.

jzaefferer:
> If the name attribute is truly required, which is not HTML5 purist friendly, then some kind of error should be thrown into the console to alert the developer as to why his code is failing.

For now it is, though there is at least one ticket to investigate changing that. Could you update your PR to throw a decent error message, as you describe? That would help. Thanks!

PR #717, opened 2013-05-13
Allow validation across all inputs with same name

imdahmd:
The following piece of code in jquery.validate.js, `elements()` function, explicitly ignores all elements with the same name (except for the first)

``` javascript
...
// select only the first element for each name, and only those with rules specified
if ( this.name in rulesCache || !validator.objectLength($(this).rules()) ) {
  return false;
}
...
```

IMO, we should rather apply the same validation rules for the elements with the same name. For which the modified piece of code would be:

``` javascript
elements: function() {
    var validator = this,
        rulesCache = {};

    // select all valid inputs inside the form (no submit or reset buttons)
    return $(this.currentForm)
    .find("input, select, textarea")
    .not(":submit, :reset, :image, [disabled]")
    .not( this.settings.ignore )
    .filter(function() {
        if ( !this.name && validator.settings.debug && window.console ) {
            console.error( "%o has no name assigned", this);
        }

        // select this element if this has the same name as one in cache (may be was dynamically added)
        if ( this.name in rulesCache ) {
            if ( !validator.objectLength($(this).rules()) ) {
                $(this).rules('add', rulesCache[this.name]);
            }
            return true;
        }

        // select only the element with rules specified
        if ( !validator.objectLength($(this).rules()) ) {
            return false;
        }

        // Add rules to the cache
        rulesCache[this.name] = $(this).rules();
        return true;
    });
},
```

What this also makes possible is, if inputs are dynamically added (in case where multi-inputs exist), the same validation would apply to those too.

### DIFF
--- a/jquery.validate.js
+++ b/jquery.validate.js
@@ -482,12 +482,25 @@ $.extend($.validator, {
 					console.error( "%o has no name assigned", this);
 				}
 
-				// select only the first element for each name, and only those with rules specified
-				if ( this.name in rulesCache || !validator.objectLength($(this).rules()) ) {
+				// filter out checkable elements whose name already in cache
+				if ( validator.checkable(this) && this.name in rulesCache ) {
 					return false;
 				}
 
-				rulesCache[this.name] = true;
+				// select this element (non-checkable) if this has the same name as one in cache
+				if ( this.name in rulesCache ) {
+					if ( !validator.objectLength($(this).rules()) ) {
+						$(this).rules('add', rulesCache[this.name]);
+					}
+					return true;
+				}
+
+				// select only the element with rules specified
+				if ( !validator.objectLength($(this).rules()) ) {
+					return false;
+				}
+
+				rulesCache[this.name] = $(this).rules();
 				return true;
 			});
 		},
--- a/test/rules.js
+++ b/test/rules.js
@@ -39,7 +39,8 @@ test("rules() - external", function() {
 });
 
 test("rules() - external - complete form", function() {
-	expect(1);
+	// twice because #form has two inputs with the name 'action'
+	expect(2);
 
 	var methods = $.extend({}, $.validator.methods);
 	var messages = $.extend({}, $.validator.messages);


jtmkrueger:
The big issue with this appears to be that it breaks on dropdowns. I get `TypeError: element is undefined` errors on any duplicate dropdowns.

naozoyoshioka:
I think elements method can be updated to support validating elements having a same name.

Here is the current elements method.
```javascript
		elements: function() {
			var validator = this,
				rulesCache = {};

			// select all valid inputs inside the form (no submit or reset buttons)
			return $( this.currentForm )
			.find( "input, select, textarea" )
			.not( ":submit, :reset, :image, [disabled], [readonly]" )
			.not( this.settings.ignore )
			.filter( function() {
				if ( !this.name && validator.settings.debug && window.console ) {
					console.error( "%o has no name assigned", this );
				}

				// select only the first element for each name, and only those with rules specified
				if ( this.name in rulesCache || !validator.objectLength( $( this ).rules() ) ) {
					return false;
				}

				rulesCache[ this.name ] = true;
				return true;
			});
		},
```
I was wondering why elements having the same name are removed in the below method.
Then I commented out
```javascript
				rulesCache[ this.name ] = true;
```
and test the below HTML.
```
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="utf-8">
	<title>jQuery validation plug-in - same name</title>
	<script src="../lib/jquery.js"></script>
	<script src="../dist/jquery.validate.js"></script>
	<script>
	$(function() {
		$("form").validate();
	});
	</script>
</head>
<body>
	<form method="get" action="">
		<label>Texts</label>
		<input name="text" type="text" required>
		<input name="text" type="text" required>
		<br/>
		<label>Radio</label>
		<input name="radio" type="radio" required>Radio 1
		<input name="radio" type="radio" required>Radio 2
		<br/>
		<label>Checkbox</label>
		<input name="checkbox" type="checkbox" required>Checkbox 1
		<input name="checkbox" type="checkbox" required>Checkbox 2
		<br/>
		<input type="submit" value="Submit">
	</form>
</body>
</html>
```
When clicking Submit button, suprisingly the result is same as the one before commenting out.
That is, only the first element of each text, radio, and checkbox shows "This field is required.".
So I guess the reason why elements having the same name are removed in elements method is to save time by omitting the processing which doesn't affect the result.

But if we put id attributes, the situation changes. I modifies the above HTML to the below.
```
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="utf-8">
	<title>jQuery validation plug-in - same name</title>
	<script src="../lib/jquery.js"></script>
	<script src="../dist/jquery.validate.js"></script>
	<script>
	$(function() {
		$("form").validate();
	});
	</script>
</head>
<body>
	<form method="get" action="">
		<label>Texts</label>
		<input id="name1" name="text" type="text" required>
		<input id="name2" name="text" type="text" required>
		<br/>
		<label>Radio</label>
		<input id="radio1" name="radio" type="radio" required>Radio 1
		<input id="radio2" name="radio" type="radio" required>Radio 2
		<br/>
		<label>Checkbox</label>
		<input id="checkbox1" name="checkbox" type="checkbox" required>Checkbox 1
		<input id="checkbox2" name="checkbox" type="checkbox" required>Checkbox 2
		<br/>
		<input type="submit" value="Submit">
	</form>
</body>
</html>
```
If the code has
```javascript
				rulesCache[ this.name ] = true;
```
the result is same.
But if I comment it out, both text fields show "This field is required.".

I think if all elements having a same name are validated, it's better as the specification of jQuey Validation Plugin and it's OK even if application developers have to put id attributes to do it.

So I think the code to validate all elements having a same name is below.
```javascript
		elements: function() {
			var validator = this,
				rulesCache = {};

			// select all valid inputs inside the form (no submit or reset buttons)
			return $( this.currentForm )
			.find( "input, select, textarea" )
			.not( ":submit, :reset, :image, [disabled], [readonly]" )
			.not( this.settings.ignore )
			.filter( function() {
				if ( !this.name && validator.settings.debug && window.console ) {
					console.error( "%o has no name assigned", this );
				}

				if (this.id) {
					if ( !validator.objectLength( $( this ).rules() ) ) {
						return false;
					}
				} else {
					// select only the first element for each name, and only those with rules specified
					if ( this.name in rulesCache || !validator.objectLength( $( this ).rules() ) ) {
						return false;
					}

					rulesCache[ this.name ] = true;
				}
				return true;
			});
		},
```

zroadhouse-wsm:
Ran into this yesterday. Found a fairly simple patch here => http://blog.kyawzinwin.me/jquery-validation-for-array-of-input-elements/

Gist:
https://gist.github.com/zroadhouse-wsm/63dacd7df7b4d987adbe

In my case, all my inputs have unique ids but the names are the same to allow for array values on the server side.

jptinsman:
I was wondering if this was ever going to be merged. Last I saw was a discussion board saying that this PR is preferred as it is more robust.  I have 2 locations where I have n check boxes and at least 1 must be checked, the number of check boxes varies and cannot have static names.

staabm:
This changes some very specific cases within the core so I am very carefull about merging it, because it migt break stuff.

@Arkni whats your feelings about this change? Any BC implications?

At the very least we need to rebase the suggested patch to see if it does not break the current tests.

Arkni:
@staabm I will investigate this soon.

leeyh20:
What happened to this PR?

leeyh20:
Will this PR be merged?

PR #998, opened 2014-01-30
The event triggering the validation (if any) is now being sent to each v...

TweeZz:
The event triggering the validation (if any) is now being sent to each validation method.
I did not find any other way to achieve what I need (see http://stackoverflow.com/questions/21450350/how-to-temporarily-not-validate-on-keyup-for-1-specific-html-element).

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -249,26 +249,26 @@ $.extend($.validator, {
 				this.addWrapper(this.errorsFor(element)).hide();
 			}
 		},
-		onfocusout: function( element ) {
+		onfocusout: function( element, event ) {
 			if ( !this.checkable(element) && (element.name in this.submitted || !this.optional(element)) ) {
-				this.element(element);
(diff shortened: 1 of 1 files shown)


staabm:

TweeZz:
Hi,

No one was able to come up with a different solution to solve the problem I explained on stackoverflow. We are using the changes I made in our application.
So yes, I still need it :)

Arkni:
@TweeZz 

Can you describe your solution, maybe I can help you with something?

I mean by your solution the one you mentioned in your [answer](http://stackoverflow.com/a/21457220) on StackOverflow.

TweeZz:
I don't understand your question. I did not want to trigger validation for some element depending on the type of event.
To be able to check the event type, I had to pass it in as a parameter and to be able to achieve that, I made the code changes you can find in my pull request.

PR #1003, opened 2014-02-01
Quickfix for issue #1002.

msimonides:
Adding the field name to the `submitted` map causes it to be considered for processing in the default `onkeyup` event handler.
Although to me it seems that calling `formatAndAdd` here could be better.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1223,6 +1223,7 @@ $.extend($.validator, {
 						errors[element.name] = previous.message = $.isFunction(message) ? message(value) : message;
 						validator.invalid[element.name] = true;
 						validator.showErrors(errors);
+						validator.submitted[element.name] = message;
 					}
 					previous.valid = valid;
 					validator.stopRequest(element, valid);


Arkni:
I followed the steps you provided in #1002 and can't reproduce the issue you reported.

Here is a jsfiddle that works using _v1.14.0_: http://jsfiddle.net/Arkni/2pq4166j/

You can test it, maybe I missed something.

Arkni:
@msimonides 
Any response to my above comment ?

msimonides:
@Arkni: thanks for looking into it.
I'm sure the problem was visible when the processing took noticeable time. I have modified your fiddle accordingly (and also switched to using jquery.validation 1.11.1) but couldn't see anything obviously wrong. And I don't remember the exact issue at this point, it's been almost two years.

I suggest looking at it by asking this question:
why triggering an error with other methods (e.g. `minlength`) results in adding the field to the `submitted` map (via a call to `formatAndAdd()`), yet the `remote` method doesn't?
Or why doesn't `formatAndAdd()` get called when the `remote` method triggers a validation error?

If these questions seem valid to you then it may be worth figuring this out.
Otherwise please just reject this pull request as I am unable to provide any more details on what I considered a bug in this scenario.
Sorry about that.

Arkni:
> I suggest looking at it by asking this question:
> why triggering an error with other methods (e.g. minlength) results in adding the field to the submitted map (via a call to formatAndAdd()), yet the remote method doesn't?
> Or why doesn't formatAndAdd() get called when the remote method triggers a validation error?

Good question, I also was wondering about that when browsing the source code. The answer I found was that the `remote` rule can use a message returned from the backend of you application.
What I mean is, instead of returning `false` as a response of a request, you can also return an error `message` which indicate that the element is invalid.

The `formatAndAdd` method wasn't designed in that aspect to accept a message as a param, only the `element` and the attached `rule` are the arguments. Using those informations, `formatAndAdd` call the method `defaultMessage` to get one of the predefined error messages from:
* The custom message you use when calling `.validate()`
* The attributes `data-msg-x` (`x` is the name of a rule)
* The `title` attribute
* The rule's default message (bundled withing the plugin)
* Or this error message: `<strong>Warning: No message defined for " + element.name + "</strong>`

This `remote` rule use this method as well alongside the message returned from the backend (if one exist).

PR #1004, opened 2014-02-02
Don't return stale result for pending remote rules.

msimonides:
When a remote request is in flight and the remote method is invoked a second
time for the same value, the method should return "pending" instead of the
previous result.
This is achieved by tracking pending requests for each particular element and
value, not just the element.


staabm:

PR #1013, opened 2014-02-10
Fix : IE8 placeholder issue

susiii:


gavinengel:
I have this issue as well.  I'm attempting to get this solution in place: http://stackoverflow.com/a/9602211/1284857

Having the ability to ignore placeholder values would be a nice option to the base jqv library for browsers like IE8.

PR #1064, opened 2014-04-01
do not search the whole DOM for radio/checkbox elements

staabm:
as per comments in #1056


src/core.js
```diff
@@ -562,7 +562,7 @@ $.extend( $.validator, {
 				type = element.type;
 
 			if ( type === "radio" || type === "checkbox" ) {
-				return $( "input[name='" + element.name + "']:checked" ).val();
+				return this.findByName(element.name).find(":checked").val();
```
jzaefferer: Using `.find()` turned out to be the issue, needs to use `.filter()`.

staabm:
@jzaefferer nice, thanks for the update and sorry for not having time to fix it myself.

Issue #521, opened 2012-09-21
E-mail validation fires immediately when text is in the field.


benjohnson:
Expected: The user can type characters for an e-mail address into the field without validations firing until they leave the field. Once an error is shown, the error will clear based on keypress.

Actual: As soon as there is anything in the field, the validation errors fire. This makes the validations seem rather aggressive.

I think this is due to the fix for #244 (pull request is #396). Changing lastActive back to lastElement restores the old behaviour. Changing it back won't really work forever because of #494.

Issue #244, opened 2011-11-02
required field's onblur after initial success breaks when followed by select box


Scotchester:
Ok, so here's a wild one.

No matter what I did, I could not get my plain-jane text box with a simple required rule to throw an error without submitting the form. I stripped the page down to nothing but the single text box. It started working again. Started adding stuff back. Eventually, after enough guess-and-check, I discovered that it only stops working if the `select` box that follows it has greater than zero `option`s. (That is, an empty select box does not break it.)

[Please view this example page.](http://dev.stratejus.com/calc/broken.php)

To reproduce: enter any text into the High School Name field and tab or click away. You'll see the 'Ok!' from where I turned on Success messages to make sure it was trying to validate at all. Return to the High School Name field and delete what's there. The field should immediately report an error, but it does not, even if you changing focus.

[Now try on this page.](http://dev.stratejus.com/calc/working.php)

If you view source, you can verify that the ONLY difference between the two pages is that on the working one, I have commented out the lone `option` in the `select`. I've tested this on both Firefox 7 and Chrome 16.

I think this may be related to the following issues:
#139
#146
#220

I had several other instances of the same problem (required text boxes not failing onblur after initial success), but I was able to pin this one down very specifically to the `option` toggle, so I thought it was worth a new issue.

Thanks for any insight anyone can provide!

PR #1072, opened 2014-04-03
Core: Trigger validation when next field is already filled out.

### LINKED ISSUE #524: Validation fails to trigger when next field is already filled out
To reproduce:

On the index demo, edit the markup to fill in the second field with a valid email address. Load that demo, type "abc" into the first field, tab to the next field, tab back, delete the value.

Expected: "This field is required" error.
Actual: No error

Originally reported as #244, but its really not specific to select elements. Reverted the fix from #396 in 5934f4d4509b900594b40eb4399446dd46debf36, as the caused a regression (#521).

This seems pretty tough to fix, as it scratches on the very basics of how the plugin figures out if a field should be validated or not. Haven't yet been able to come up with a better solution.

stijnherreman:
Fixes #524 (and #244) without causing #521.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -247,7 +247,7 @@ $.extend($.validator, {
 		onkeyup: function( element, event ) {
 			if ( event.which === 9 && this.elementValue(element) === "" ) {
 				return;
-			} else if ( element.name in this.submitted || element === this.lastElement ) {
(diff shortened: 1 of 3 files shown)


jzaefferer:
Closing due to inactivity.

PR #1201, opened 2014-07-06
core: Fix incorrect regexp

### LINKED ISSUE #1200: Bug in showLabel method - creates duplicate aria-describedby elements
Found in 1.13.0

This code (I assume) is trying to send the \b to the regex, but its being escaped by the JS

Should be something like "\\b" + errorID + "\\b"?

CODE:
} else if ( !describedBy.match( new RegExp( "\b" + errorID + "\b" ) ) ) {
                        // Add to end of list if not already present
                        describedBy += " " + errorID;
                    }

This creates loads of duplicate aria-describedby in the markup

tractorcow:
Fixes #1200

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -770,7 +770,7 @@ $.extend( $.validator, {
 					// Respect existing non-error aria-describedby
 					if ( !describedBy ) {
 						describedBy = errorID;
-					} else if ( !describedBy.match( new RegExp( "\b" + errorID + "\b" ) ) ) {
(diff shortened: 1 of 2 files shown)


rwil02:
NOT fixed in 1.13.1 (from nugget)

If using (for example) an id like "avd.Enabled", it fails to match and adds forever

omuleanu:
looks fixed for me in 1.14

PR #1211, opened 2014-07-17
Ignore rules with null or undefined

edgarjs:
Just like it ignores explicit `false`, it should ignores `null` and `undefined` since those don't respond to `.param` neither.

Error: http://jsfiddle.net/MYLuL/


staabm:
Closing, as very old. @edgarjs if you still need this, please shout out.

PR #1221, opened 2014-07-24
Pending class

alokito:
This pull request adds a "pending" class to the input element while there is remote validation pending. This allows you, for example, to add a loading animation to visually indicate that there outstanding activity for the field.

The new unit test passes in IE, safari, chrome and firefox. There is a unit test failing in IE, but it is not one I wrote, and it was already failing in my version of IE before I added my code:

    validator: focusInvalid() (1, 0, 1)
    29 ms1.Expected at least one assertion, but none were run - call expect(0) to accept zero assertions.


staabm:
@Arkni any comments on this?
@alokito any chanche you could rebase it against master?

src/core.js
```diff
@@ -868,6 +869,7 @@ $.extend( $.validator, {
 		startRequest: function( element ) {
 			if ( !this.pending[ element.name ] ) {
 				this.pendingRequest++;
+				$(element).addClass(this.settings.pendingClass);
```
Arkni: Please add one space before `(` and `)`

See http://contribute.jquery.org/style-guide/js/#spacing

src/core.js
```diff
@@ -879,6 +881,7 @@ $.extend( $.validator, {
 				this.pendingRequest = 0;
 			}
 			delete this.pending[ element.name ];
+			$(element).removeClass(this.settings.pendingClass);
```
Arkni: Same as above

test/methods.js
```diff
@@ -400,6 +400,32 @@ asyncTest("remote", function() {
...
+asyncTest("remote, pending class added to element while call outstanding", function() {
+	expect(3);
+	var e = $("#username"),
+		v = $("#userForm").validate({
+			rules: {
+				username: {
+					remote: "users.php"
+				}
+			},
+			submitHandler: function() {
```
staabm: This should not be part of is test, right.

test/methods.js
```diff
@@ -400,6 +400,32 @@ asyncTest("remote", function() {
...
+				ok( false, "submitHandler may never be called when validating only elements");
+			}
+		});
+
+	$(document).ajaxStop(function() {
```
staabm: Instead of tis global `ajaxStop` hack, we should better pass a success callback into the remote method.

You can pass any option to the remote method which jquery supports for `$.ajax`, see example at http://jqueryvalidation.org/remote-method/

test/methods.js
Arkni: This should be removed from the test.

test/methods.js
Arkni: See also: https://github.com/jzaefferer/jquery-validation/blob/master/test/methods.js#L415-L430

test/methods.js
```diff
@@ -400,6 +400,32 @@ asyncTest("remote", function() {
 	strictEqual( v.element(e), true, "still invalid, because remote validation must block until it returns; dependency-mismatch considered as valid though" );
 });
 
+asyncTest("remote, pending class added to element while call outstanding", function() {
```
Arkni: I forgot to mention the spacing issue in this test as well.
Please fix them as I mentioned above

Arkni:
@staabm I can rebase the PR on top of master and adjust it per above feedback.
of course the attribution and the authorship should still preserved.

I you want that, I'll open a new PR for that.

alokito:
Sorry I lost track of this, I should have some time to look at this later
today.

Best,
Alok

Arkni:
@alokito thanks :)

test/methods.js
alokito: success callback didn't work, because it gets invoked before the pending class is removed. However, the complete callback worked fine.

alokito:
So... I merged the trunk into my pull request branch. I'm not sure that was a great idea. I could easily rebase my changes onto master and submit a new pull request, let me know.

test/methods.js
```diff
@@ -396,6 +396,29 @@ asyncTest( "remote", function() {
...
+						url: "users.php",
+						complete: function() {
+							strictEqual( e.hasClass( "pending" ), false, "not pending since ajax call complete" );
+							start();
+						}
+					}
+				}
+			}
+		} );
+	strictEqual( e.hasClass("pending"), false, "not pending since no data entered" );
```
Arkni: Please add one space:
```js
                        ↓       ↓
strictEqual( e.hasClass("pending"), false, "not pending since no data entered" );
```

test/methods.js
```diff
@@ -396,6 +396,29 @@ asyncTest( "remote", function() {
...
+	e.val( "Peter" );
+	// this fires off the validation:
+	v.element( e );
+	strictEqual( e.hasClass( "pending" ), true, "pending while validation outstanding" );
+});
```
Arkni: here as well.
=> `) }`

Arkni:
Otherwise, it LGTM :)

After adjuster as per feedback, Could you please squash your commits into one.

Thanks :)

alokito:
Should I submit another pull request with the squashed commits, or is there another way?

Arkni:
> Should I submit another pull request with the squashed commits, or is there another way?

The best option IMHO (and should be easier for you), is to create a new clean branch from the master (upstream), and include your change there, then send a new pull request.

alokito:
Sounds good. I should probably do two commits, one for the changes to Core and one for Test, correct?

Arkni:
> I should probably do two commits, one for the changes to Core and one for Test, correct?

Sounds good.

Arkni:
Please close this PR in favor of the new one.

Issue #1220, opened 2014-07-24
.isValid() method, without marking the element


davesierra:
I need a way to check if an element is valid while only returning true/false, _without_ marking the element with the error class. Similar to, but also unlike .valid() which will actually mark the element with the classname valid/error and update the message. 

Does anyone have an elegant solution for this? This is what I came up with for now:

```
isValid: function() {
    var validator = $( this[ 0 ].form ).validate();
    var cleanElement = validator.clean( this ),
        checkElement = validator.validationTargetFor( cleanElement ),
        result = true;

    result = validator.check( checkElement ) !== false;

    return result;
}
```

The reason for this is that on my onkeyup function, I wanted to delay the validation by a second or two before it got validated again. 

For example:
I have two rules for email, one for required, and one for valid email address. If user tries to submit without a value, the user originally gets the err message is "required". Once focused and they start inputting, even with one key, the err message immediately changes to "not valid". Obviously not UX friendly, the delay's goal is to allow the user enough time to input their correct email.

```
onkeyup: function (element, event) {
...
clearTimeout(timeout);
if ( $(element).hasClass('error') && event.keyCode !== 9 ) {
    // if the element is valid, validate it as normal 
    // .valid() would of validated, we want to check only if its valid first
    if ( $(element).isValid() ) {
        $element.valid();
    } 
    // else the element is not valid, give the user a delay on input 
    // before returning another error
    else {
        // On each keyup, give user a second to enter something
        timeout = setTimeout(function () {
            console.log('give it a sec')
            $element.valid();
        }, 1000);
    }
}
}
```

jcspader:
+1

yanickrochon:
BTW: how this issue can be a duplicate of a later one? :stuck_out_tongue_closed_eyes:

joneshwangoo:
Haha LOL, duplicate one of later issue.

ShaneHudson:
> 
> 
> there is a `check(element)` method which would help you, but it seems it is not part of the public API
> https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L582

For anyone that finds this page through Google like I did, this is the answer (I missed it when I first saw the page because of the comments after it about duplicates). Check is not in documentation but it does work.

```
function isValid(el) {
    return $("form").validate().check(el);
}
```

PR #1270, opened 2014-09-15
Core: Adding ability to not show validation messages when calling valid().

tdekoning:
This feature is very usefull for when you want to check if a field is valid without showing the user he/she has done something wrong.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -86,7 +86,7 @@ $.extend($.fn, {
 		return validator;
 	},
 	// http://jqueryvalidation.org/valid/
-	valid: function() {
+	valid: function( showErrors ) {
 		var valid, validator;
 
 		if ( $( this[ 0 ] ).is( "form" ) ) {
@@ -95,7 +95,7 @@ $.extend($.fn, {
 			valid = true;
 			validator = $( this[ 0 ].form ).validate();
 			this.each( function() {
-				valid = validator.element( this ) && valid;
+				valid = validator.element( this, showErrors ) && valid;
 			});
 		}
 		return valid;
@@ -387,7 +387,7 @@ $.extend( $.validator, {
 		},
 
 		// http://jqueryvalidation.org/Validator.element/
-		element: function( element ) {
+		element: function( element, showErrors ) {
 			var cleanElement = this.clean( element ),
 				checkElement = this.validationTargetFor( cleanElement ),
 				result = true;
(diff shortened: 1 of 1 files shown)


jcspader:
+1 
I created a function with some fields. In onChange event, I test those fields, and if is valid, it call other  function.
I searched and I found issue  #1220. I think is a good idea.
I don't like @staabm suggestion. I think is function the library give information, if is valid or not.

Nettsentrisk:
It's a bit of a pain that valid() applies a call to validate(), meaning there's no way to easily check whether the form is valid after it has already been validated. I guess the only way to accomplish that right now is to access the validator.invalid object on the jQuery data attached to the form?

My use case: I'm hooking onto the submit event from some other script on a form that's being validated through jQuery validation. Now, I want to know within that code whether the form that's being submitted is valid or not.

If I call form.valid(), it will redo the validation that just happened due to the submit event, so it's doing the validation logic twice in a row just to check if it's valid...

I guess I would have to hook up a submitHandler to this instance of the form, but can I set that after the validate-setup has already been run? Hmm.

Arkni:
> I guess the only way to accomplish that right now is to access the validator.invalid object on the jQuery data attached to the form?

You shouldn't use internal variables, those may change from version to another which will cause your code to break. As an alternative, there are some internal methods that you can use:
* `validator.numberOfInvalids()`: return the count of invalid elements in your form (only the validated ones).
* `validator.invalidElements()`: returns a jQuery array that contains the invalid elements (only the validated ones).

> I guess I would have to hook up a submitHandler to this instance of the form, but can I set that after the validate-setup has already been run? Hmm.

From [submitHandler's docs](http://jqueryvalidation.org/validate/#submithandler):
```
submitHandler (default: native form submit)
   Type: Function()
   Callback for handling the actual submit when the form is valid. Gets the form as the
   only argument. Replaces the default submit. The right place to submit a form via
   Ajax after it is validated.
```

Nettsentrisk:
Yes, but can you set up a new submitHandler after the form has already been set up with a call to .validate()? Will .validate(options: { submitHandler: function() { new submithandler } }) only overwrite the previous submitHandler, or will it set all the other options back to default also?

For now I've solved this by setting a data attribute on the form and checking if it exists when I first set up the form with .validate() to provide the correct submitHandler, but that's not very flexible.

Arkni:
> Yes, but can you set up a new submitHandler after the form has already been set up with a call to .validate()? Will .validate(options: { submitHandler: function() { new submithandler } }) only overwrite the previous submitHandler, or will it set all the other options back to default also?

The second call of `.validate( {} )` will be ignored by the plugin.
But I guess you can overwrite the submitHandler this way:
```js
var validator = $( "yourForm" ).data( "validator" );
validator.settings.submitHandler = function() {
    // new submitHandler
};
```

PR #1310, opened 2014-10-24
New method: Brazilian CPF number

gorork:
New method that validates Brazilian CPF numbers.


4devs-net-br:
https://4devs.net.br/gerador-cpf

PR #1330, opened 2014-11-20
validate multiple elements by same name

monotom:
see http://stackoverflow.com/questions/931687/using-jquery-validate-plugin-to-validate-multiple-form-fields-with-identical-nam/4136430#4136430

tests updated and checked, they cover the problem if you dont expect 1 validation per name rule


staabm:
Closing in favor of https://github.com/jzaefferer/jquery-validation/pull/717.

Thanks for the PR though

PR #1353, opened 2014-12-11
Allow Customer ResponseTransformers on Remote Validation

clamp:
@jzaefferer This will keep existing functionality, but also now allow for a user to provide a param to the remote called `responseTransformer`. Unlike jQuery.ajax()'s `dataFilter` this will now allow for a user to handle errors (404s, 500s) etc properly and display valid or invalid on it.

Moving the logic to handle these errors to ajaxComplete() is useful because cross-domain request will not trigger an error. 

The function provided to `responseTransformer` must return `true` or `false`. This is because that response is then passed to the shared logic that renders the form invalid with an error message or allows the form to be submit.

Let me know if you have any suggestions/questions. I think this will benefit.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,1307 +1,1327 @@
 $.extend($.fn, {
-	// http://jqueryvalidation.org/validate/
-	validate: function( options ) {
+    // http://jqueryvalidation.org/validate/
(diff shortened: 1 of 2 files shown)


staabm:

PR #1361, opened 2014-12-21
Fixed number validation error. Fixes #1359

### LINKED ISSUE #1359: Number validation Error
Hi, the number rule validation has a bug, if I simply type a - character it pass the validation.

greetings,

wozzo:
Changed line 1153 to include a lookaround for a digit in the event a minus sign has been matched so that a "-" on it's own does not match.

Added (?=\d) after the -?


jzaefferer:
Thanks for the contribution! I've reworked this quite a bit, extending the existing test instead of adding a new one.

PR #1373, opened 2015-01-14
Additionals: fixed the accept method regex, regex should be escaped

### LINKED ISSUE #1243: "accept" validator not working on application/epub+zip mimetype
A validation rule like the following will fail validation even if the mimetype is the correct one:

```
    'epub_file':
    {
      required: true,
      accept: "application/epub+zip"
    }
```

The file I used to test is Orwell's "Animal Farm" from https://www.epubbooks.com/book/1074/animal-farm

After selecting the file and validation failed I ran the following:

```
document.getElementById('input-epub-upload').files[0]['type']
```

The output was:

```
"application/epub+zip"
```

mishal:
Fixes #1243


jzaefferer:
That looks decent. Could you add unit tests for this method? Even if they need to be wrapped in a feature detect for `element.files`, that would still be a big improvement over no coverage at all.

mishal:
I tried to create test for it, but I'm not sure how to mock the file input functionality, since JS cannot update the value if the file input... 

Do you have an idea how to test it?

jzaefferer:
You could start with an input of type file, select that from the DOM, then overwrite the `files` property. Not sure if its writable, but if it is, this might work (not tested):

```html
<input type="file" id="accept-input">
```
```js
var element = $( "#accept-input" )[0];
// Assign something here for each assertion, not sure what the right format is
element.files = [ "myfile.docx" ];
ok( $.validator.method.accept( null, element, "docx" ), "something something is valid" );
```

I have no idea what the right format for `files` and `param` are, but I hope this helps. Thanks for looking into it!

mishal:
Hmmm, the input property `files` is readonly for security reasons.

jzaefferer:
Makes sense. Maybe we can mock the DOM element from scratch? Along with the `files` property it needs to be pass the `this.optional()` check and respond to `$(element).attr("type")` with `"file"`.

mishal:
I had the same idea.. I will try it :)

mishal:
I've just created the tests, the last one is for invalid mime type. I'm not sure if the last one check for `dependency-mismatch` is ok...

test/test.js
```diff
@@ -1771,3 +1771,98 @@ test( "Validation triggered on radio and checkbox via click", function() {
...
+		length: 1,
+		item: dummy
+	},
+		input = {},
+		$form = $("<form />"),
+		proxy,
+		validator;
+
+	// dummy DOM element
+	input.type = "file";
```
jzaefferer: You can define these above, its fine to reference a variable defined in the same multi declaration.

test/test.js
```diff
@@ -1771,3 +1771,98 @@ test( "Validation triggered on radio and checkbox via click", function() {
...
+	input.files = fileList;
+	input.nodeName = "INPUT";
+
+	validator = new $.validator({}, $form[0]),
+	proxy = $.proxy($.validator.methods.accept, validator, null, input, "image/*");
```
jzaefferer: This variable isn't needed, you can inline it into the assertion.

test/test.js
```diff
@@ -1771,3 +1771,98 @@ test( "Validation triggered on radio and checkbox via click", function() {
...
+		proxy;
+
+	input.type = "file";
+	input.files = fileList;
+	input.nodeName = "INPUT";
+
+	validator = new $.validator({}, $form[0]),
+	proxy = $.proxy($.validator.methods.accept, validator, null, input, "application/vnd.google-earth.kml+xml");
+
+	equal(proxy(), "dependency-mismatch", "the selected file for upload is valid" );
```
jzaefferer: There's a lot of (structural) duplication between these three tests. I think a builder method that returns the necessary object(s) and takes arguments for those parts that are specific for each test would be a good way to refactor this.

Also, since this is testing a method, let's move these tests into `test/methods.js` and name the tests accordingly.

jzaefferer:
Thanks for the update! This is promising, but needs a bit of refactoring and cleanup.

mishal:
Yes, you are right, this is raw version, I will clean it up.

jzaefferer:
Never noticed the new commits until now - there are no notifications for pushes... reviewing again.

test/methods.js
```diff
@@ -11,6 +11,38 @@ function methodTest( methodName ) {
...
+	function dummy() {
+		return file;
+	}
+
+	// https://developer.mozilla.org/en-US/docs/Web/API/FileList
+	var file = {
+			name: filename,
+			size: 500001,
+			type: mimeType
+		}, fileList = {
```
jzaefferer: Add a line break after the comma

test/methods.js
```diff
@@ -11,6 +11,38 @@ function methodTest( methodName ) {
...
+			0: file,
+			length: 1,
+			item: dummy
+		},
+		input = {};
+
+	// dummy DOM element
+	input.type = "file";
```
jzaefferer: These three properties can be initialized when declaring the variable above.

test/methods.js
```diff
@@ -1282,4 +1314,25 @@ test("cpfBR", function() {
...
+			$form = $("<form />"),
+			proxy = $.proxy($.validator.methods.accept, new $.validator({}, $form[0]), null, input, "application/vnd.google-earth.kml+xml");
+	ok( proxy(), "the selected file for upload is valid" );
+});
+
+test("file accept - invalid mime type", function() {
+	var input = acceptFileDummyInput("test.kml", "foobar/vnd.google-earth.kml+xml"),
+			$form = $("<form />"),
+			proxy = $.proxy($.validator.methods.accept, new $.validator({}, $form[0]), null, input, "application/vnd.google-earth.kml+xml");
+	equal(proxy(), "dependency-mismatch", "the selected file for upload is valid" );
```
jzaefferer: The expected value doesn't match the test name ("invalid mime type"), I'd expect a `false` here (in that case, using `strictEqual` would also help).

It also looks like the message argument was just copied from previous assertions and doesn't match here.

jzaefferer:
Still a few minor issues with the tests. Please let me know when you push new commits, so that I can review them immediately.

test/methods.js
mishal: What exactly do you mean? I don't get it :(

mishal:
ping @jzaefferer

test/methods.js
jzaefferer: The return value `"dependency-mismatch"` comes from the `optional()` method, which indicates that the code in question doesn't actually, since that returns `false`. So it looks to me like the test is expecting the wrong result, and that there is an issue elsewhere causing the wrong code to run. It might be a problem with the test, it might be in the implementation.

Much less critical, the third argument, "the selected file for upload is valid", is wrong, at least based on what the test name says ("invalid mime type").

jzaefferer:
Hope thats more clear. I'll be away the next few days, will check again next week.

mishal:
Sorry for the delay, I had urgent work to do.

ping @jzaefferer

mishal:
@jzaefferer did you have a chance to see this?

jzaefferer:
Sorry for the delay. Could you rebase this against master? Thanks.

twolfson:
@mishal As @jzaefferer requested in #1512, I have squashed/rebased this PR into #1531. Authorship and the original diff should be maintained but now in 1 commit and on top of the latest `master`.

mishal:
Thanks, closing, please merge #1531

PR #1379, opened 2015-01-16
Proposed fix for #1378 - Added parseFloat before value and param are com...

### LINKED ISSUE #1378: param & value for min & max rules should be type cast when rules are specified through data attributes
When rules are specified through data attributes such as data-rule-min="77.50", before param and value can be compared they should be typecast, such as through parseFloat. Currently, both value and param are strings. As a result, string comparison is performed instead.

Currently, the type of both value and param are string as tested in:

    // http://jqueryvalidation.org/min-method/
	min: function( value, element, param ) {
		console.log( typeof value, typeof param ); // console ouput: string string
		return this.optional( element ) || value >= param;
	},

This is reproduced in this jsfiddle: http://jsfiddle.net/1rxvgL4v/13/

If you entered 123 into the amount field which is (obviously) larger than 77.50, the error message appears.

If you entered 8 which is less than 77.50, the error message disappears.

zhenrong:
Fix #1378 - value and param variables should be typecast before
comparison. This is required if rules are specified through
data-attributes

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1216,12 +1216,12 @@ $.extend( $.validator, {
 
 		// http://jqueryvalidation.org/min-method/
 		min: function( value, element, param ) {
-			return this.optional( element ) || value >= param;
+			return this.optional( element ) || parseFloat( value ) >= parseFloat( param );
 		},
 
 		// http://jqueryvalidation.org/max-method/
 		max: function( value, element, param ) {
-			return this.optional( element ) || value <= param;
+			return this.optional( element ) || parseFloat( value ) <= parseFloat( param );
 		},
 
 		// http://jqueryvalidation.org/range-method/


jzaefferer:
Why did you close this?

zhenrong:
Hi Jörn,

I closed the pull request because i realized it didnt pass all the tests. My apologies for that...  I realized also that my solution will not work if date and string comparison is required.

ZR

jzaefferer:
Okay, thanks for letting me know. If you can find a solution that solves the problem you reported, let me know, another PR is definitely welcome.

zhenrong:
Hi 

Thanks for the encouragement.. I'm new to GitHub and i'm sorry for creating lots of unwanted commits...

Here's my latest PR. This time round, it fixes the issue reported and passes the travis test.
https://github.com/jzaefferer/jquery-validation/pull/1391

ZR

Issue #1283, opened 2014-10-02
remote method not working as expected


Adspectus:
I try to validate a form with one input field which should be checked remotely. On the very first invocation, the check is done but the return code of the `.valid()` method is true regardless of the result of the remote check. In subsequent checks, the return code is false if the remote check failed. I might believe that this is due to the asynchronous nature of the remote check, but this would make remote checks completely senseless. Since i do not believe this, my question is how to do remote checks in a way such the return code of validation reflects the outcome of the remote check (in addition to the other checks of course)?

cravecode:
I love this plugin and it's used in a Drupal contrib module I depend heavily on. The non-blocking nature of a remote validator is pretty problematic. Instead of complaining (not saying others are, it's just my normal nature), I'd like to help however I can. 

Here is what I'm thinking. I plan on implementing a custom validator that mimics the `remote` validator called `blockingRemote`. It essentially does the same thing as the `remote` validator but puts a "state" flag on the element perhaps as a `data` value or CSS class. I don't know if the "state" validation should be a separate validator to leverage unique messages or not?

I as can tell, the different "states" of a `blockingRemote` validation are:
  * unvalidated
  * pending
  * failed
  * validated
  
I'm not stuck on the naming convention, feel free to critique. 
Is this a ridiculous approach? I'm sure there has been a lot of thought into a solution. How can I help?

jzaefferer:
I definitely welcome any help I can get with improving the remote method. A look at the label, which is probably still missing some issues, also documents that there are plenty of issues: https://github.com/jzaefferer/jquery-validation/labels/Remote

I don't really have a solution, otherwise I would have implemented it by now ;) That said, I need some details for your proposal to understand its merits. Specifically, what does the `state` flag accomplish? How is it used, how does the blocking look like? Would there be any changes affecting the user (like disabling the field while its being validated)? If not, can you make this a patch to the existing remote method? Obviously it shouldn't break existing applications and the more existing issues it can address, the better.

PS: I'll be away the next few days, but will read any proposals you might write next week and going forward.

jzaefferer:
I'm sorry for the lack of activity on this issue. Instead of leaving it open any longer, I decided to close old issues without trying to address them, to longer give the false impression that it will get addressed eventually, especially after several years with no activity. It doesn't mean I'm abandoning the project, just that I'm unable to work through 200+ open issues with the little time I can afford to spend on this project.

[...]

PR #1380, opened 2015-01-19
Solution that fixes remote async calls from form.valid()

danieloprado:


jzaefferer:
Thank you for the contribution! This looks like its supposed to fix #1283. Its an interesting idea, though I suspect there will be many problems involved with this. For a start, could you add a test that covers the new signature?

jzaefferer:
Closing due to inactivity.

PR #1384, opened 2015-01-28
Core: Skip validation when settings is empty

wmh:
My case is using jquery-validation with [jquery-upload-file plugin](https://github.com/hayageek/jquery-upload-file), these's a dynamic created form inside the original form and it's not able to get the correct settings. The solution is to skip the validation if the settings not found.
reproduce the issue: http://jsbin.com/macixayoyo/3/


jzaefferer:
Thanks for the contribution. This looks like it might have unintended side-effects. For a start, could you add a test that covers your usecase?

jzaefferer:
Closing due to inactivity.

Issue #1331, opened 2014-11-20
aria-describedby character escaping (with fix suggstion)


a-barry:
ref #1269 

When using non label error elements with asp.net web forms the $ character is used in element IDs/names. This causes an error from sizzlejs on the second form submit (dues to the $ in the aria-describedby value).

core.js line 773 needs $ adding to the escape characters ;

``` javascript
- errorID = error.attr( "id" ).replace( /(:|\.|\[|\])/g, "\\$1");
+ errorID = error.attr( "id" ).replace( /(:|\.|\[|\]|\$)/g, "\\$1");
'''
```

pascallapointe:
Error Id in aria-describedby that contain escape character (\) are duplicate each time the field is validate.
Those escape characters need to be escape too.

Try this on line 788 in core.js:
`!describedBy.match( new RegExp( "\\b" + errorID.replace( /(\\)/g, "\\\\$1" ) + "\\b" ) )`

PR #1388, opened 2015-02-01
Fix for duplication of existing error aria-describedby. Issue 1331.

pascallapointe:
#1331 bottom comment.

### DIFF
--- a/test/error-placement.js
+++ b/test/error-placement.js
@@ -311,6 +311,21 @@ test( "test existing non-error aria-describedby", function( assert ) {
 	strictEqual( "", $("#testForm17text-error").text(), "Error label is empty for valid field" );
 });
 
+test( "test existing non-error aria-describedby with bracket", function( ) {
+	expect( 2 );
+	var form = $( "#testForm18" ),
+		field = $( "#testForm18\\[text\\]");
+
+	field.attr("aria-describedby", "testForm18\\[text\\]-description testForm18\\[text\\]-error");
+
+	form.validate({
+		errorElement: "span"
+	});
+
+	ok( !field.valid() );
+	strictEqual( field.attr("aria-describedby"), "testForm18\\[text\\]-description testForm18\\[text\\]-error", "aria-describedby equal to 'testForm18\\[text\\]-description testForm18\\[text\\]-error'" );
+});
+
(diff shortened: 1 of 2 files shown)


staabm:
please add a unit test

pascallapointe:
Yes, my bad, I'll do it some time this week.

jzaefferer:
@pascallapointe would still be great if you could add the unit test.

pascallapointe:
Unit tests added. Let me know if something is missing.

test/error-placement.js
```diff
@@ -311,6 +311,21 @@ test( "test existing non-error aria-describedby", function( assert ) {
 	strictEqual( "", $("#testForm17text-error").text(), "Error label is empty for valid field" );
 });
 
+test( "test existing non-error aria-describedby with bracket", function( ) {
+	expect( 2 );
+	var form = $( "#testForm18" ),
+		field = $( "#testForm18\\[text\\]");
+
+	field.attr("aria-describedby", "testForm18\\[text\\]-description testForm18\\[text\\]-error");
```
jzaefferer: Why are you setting this to also reference the "-error"? Shouldn't that get added by the plugin? With this in place you're only testing that the plugin doesn't remove anything, not that it correctly adds something.

pascallapointe: I did it to simulate a second validation pass, should have use validator.element() after the form.validate().
ie: Check/Uncheck a checkbox after the initial submit validation have been done will trigger validator.element(), previous error reference could not be found and was added again and again.

The minor fix I provide is to avoid the 'xxx-error' reference to be add multiple time, I actually testing that the reference isn't added again if already there.

jzaefferer: I still don't understand this approach. You can call `.valid()` multiple times to verify the change isn't applied multiple times. That way you can also verify that it works properly in the first place.

pascallapointe: I understand your position and your right.
I did it that way because using .valid() do not reproduce the problem I got by checking/unchecking a required checkbox. .element() don't reproduce it neither. After many hours looking around I had just given up. I'm sorry but I will have to dig deeper into the plugin to understand why I'm not able to simply reproduce it with common method.

pascallapointe:
Close until I got better unit test.

PR #1391, opened 2015-02-03
Core: dataRules method to convert value to number for number inputs

zhenrong:
Updated dataRules method to convert value to number for number inputs.
Proposed fix for #1378

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -996,12 +996,26 @@ $.extend( $.validator, {
 	},
 
 	dataRules: function( element ) {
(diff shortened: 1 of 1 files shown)


staabm:
Hi! thanks for the PR. could you also add a unit test?

jzaefferer:
This also duplicates a lot of the `attributeRules` method. Needs a refactoring to reuse that code instead.

jzaefferer:
I did the refactoring and added a test.

PR #1392, opened 2015-02-03
Core: Added destroy method to validate

Westie:
Hi!

I've created a method that will reset jQuery validate and put it in the state where the object can be re-initiated.


staabm:
Hi! whats the use-case you have in mind? this also requires unit tests to land.

src/core.js
```diff
@@ -913,6 +913,15 @@ $.extend( $.validator, {
...
 				message: this.defaultMessage( element, "remote" )
 			});
+		},
+
+		// cleans up all forms and elements, removes validator-specific events
+		destroy: function() {
+			this.resetForm();
+
+			$( this.currentForm ).unbind( "submit.validate" );
+			$( this.currentForm ).find(" :input ").unbind( "focusin.validate focusout.validate keyup.validate click.validate invalid-form.validate" );
```
jzaefferer: None of these events are bound to any inputs directly, they all use event delegation.

jzaefferer:
This looks reasonable. We could land it even without unit tests, since the relevant change is a new method. But the implementation needs to be improved, see my other comment.

Westie:
Thanks for looking over this for me. I haven't had the chance to look over this properly 'cos I just did this as a quick fix for an issue that I had encountered at work.

I can't spend too long going over this 'cos I'm at work at the moment, however, I can see that I've had a misunderstanding how validateDelegate works.

Would it be wrong of me to interpret validateDelegate as an analogue of on/bind, when using it with a selector?

jzaefferer:
Yes, `validateDelegate` is like `on` with delegation. It was created before jQuery had delegation support built-in. For some reason simply replacing it with delegate/on doesn't work. I still haven't gotten around to looking into that either.

PS: If this addresses an issue you had at work, you should ask for time to fix the issue at work, too.

Westie:
I'm surprised with that not working, that looks like an excellent use case for `on`. Was there a particular feature that wasn't working? I might have a look at it if I have time tonight.

Also, is there a specific version of jQuery that you're building for? I don't notice any particular versions in the package definition.

(also, I meant on/live instead of on/bind - thankfully I don't have to deal with the joys of $.fn.live any more!)

And tonight for me is that hour between 5 and 6pm - yes, I work until 6 at night :(

jzaefferer:
I can't remember what exactly the issue was. Since `validateDelegate` is pretty trivial, it shouldn't be hard to find out.

As for jQuery Core support: Currently its supporting down to 1.6.4, though I'm okay with dropping that and making 1.7 the minimum, to be able to use `on`. jQuery UI is doing the same with its next release. To test with various Core versions, use the links in the testsuite at the top.

Westie:
This new patch passes grunt unit tests, and all but jQuery 1.6.4 (expected) and latest on Git (unexpected) - I'll see if I can test your latest with latest.git and see if we get the same issues.

Westie:
Okay, I can confirm that there are some errors being thrown when testing with latest.git.

jzaefferer:
Good catch! Looks like `size()` is gone. I've fixed that in master. I'll rebase your PR and test with that.

src/core.js
```diff
@@ -917,6 +917,21 @@ $.extend( $.validator, {
...
 			});
+		},
+
+		// cleans up all forms and elements, removes validator-specific events
+		destroy: function() {
+			this.resetForm();
+
+			$( this.currentForm )
+				.off( "submit.validate" )
+				.off( "focusin.validate focusout.validate keyup.validate",
```
jzaefferer: This can be simplified to `".validate"` as the first argument, should do the same thing. I guess the second argument is still required, but would be good verify that.

test/test.js
```diff
@@ -1765,3 +1765,30 @@ test( "Validation triggered on radio and checkbox via click", function() {
...
 });
+
+test( "destruction of Validator", function() {
+	expect( 2 );
+
+	var form = $( "#validatorDestroy" ),
+		validate = form.validate();
+
+	form.on( "submit.validate", function( event )
+	{
```
jzaefferer: Unfortunately jscs is ignoring this - can you please stick to the existing style and remove the linebreak here?

test/test.js
```diff
@@ -1765,3 +1765,30 @@ test( "Validation triggered on radio and checkbox via click", function() {
...
+		event.preventDefault();
+
+		ok( !!$( this ).data( "validator" ), "Validator has events and object" );
+	});
+
+	form.submit();
+
+	validate.destroy();
+
+	form.on( "submit.validate", function( event )
```
jzaefferer: This doesn't make sense. You're just binding another event handler, for no good reason in the same namespace. Could run the same `ok()` assertion after `destroy()` call, no need for a submit. Also applies above, you don't need to submit the form at all to check the `data()`.

jzaefferer:
Thank you for the contribution! A bit of cleanup and refactoring, and this should be good to go.

test/test.js
Westie: Done.

jzaefferer: I'm not seeing a change here. Maybe you haven't pushed an update, yet?

test/test.js
Westie: That is indeed a good point, however, the reason I had chosen to do it this way is that the only times I had encountered errors was when I was submitting the form just after I had reset it - and some events were still active. I can change it to the way you suggested it to be if you'd like?

test/test.js
jzaefferer: PS: Updating `grunt-jscs` to latest (1.5.0) ends up with several thousand errors :/

test/test.js
jzaefferer: Without an example of the actual error you encountered, its hard to tell what was going on there. For the isolated test here, I don't see value in the submit handlers.

test/test.js
Westie: Yeah, I was typing up another message.

src/core.js
Westie: I've just confirmed this with JSFiddle, and yeah, I never expected that to work. Changes pushed and synced.

staabm:
This PR looks awesome

test/test.js
Westie: The exact error was something along the lines of 'unable to access property of undefined' or something along those lines - the :submit/:input handler was still set, and when calling `$(form).trigger("submit")` it would cause some of the old handlers to try to access the validator object, which, at this time was no longer in existence.

src/core.js
```diff
@@ -917,6 +917,15 @@ $.extend( $.validator, {
...
 				message: this.defaultMessage( element, "remote" )
 			});
+		},
+
+		// cleans up all forms and elements, removes validator-specific events
+		destroy: function() {
+			this.resetForm();
+
+			$( this.currentForm )
+				.off( "submit.validate focusin.validate focusout.validate keyup.validate click.validate" )
```
jzaefferer: So `.off( ".validate" )` won't actually work?

Westie: Aah, sorry! This was a nice little feature that I didn't know about until today. :| I'll patch and push.

jzaefferer: Okay, thanks. Just comment again on this PR after you pushed.

Westie:
There you go, done.

jzaefferer:
Thanks, that looks better. I still don't think the test should be using submit events, since the destroy method isn't related to those. Just verifying the destroy method works properly is enough.

jzaefferer:
@Westie could you look into this again? Thanks.

Westie:
Sorry for not getting back to you sooner, I had a personal project to get shipped.

I'd rather test the submit event as well, considering how it was me seeing an error being called there was what brought me down this route of adding the destroy method in the first place.

I'll commit an adapted test, testing only the dropping of data.

Westie:
Anything else you'd like me to add...?

staabm:
please add a note to the changelog about the new jquery min. version

jzaefferer:
I missed the new commit (no notifications for commits). Nevermind the changelog, I'll update that during the next release.

Reviewing the rest...

jzaefferer:
I still have a problem with the unit test. How's this instead (untested)?

```js
test( "destroy()", function() {
	expect( 2 );

	var form = $( "#validatorDestroy" ),
		validate = form.validate();

	strictEqual( $( this ).data( "validator" ), validate );

	validate.destroy();
	strictEqual( $( this ).data( "validator" ), undefined );
});
```
Could be extended to check that `resetForm` runs and event handlers are removed, since that's what `destroy()` actually does. Note that I've also updated the name of test.

test/index.html
```diff
@@ -370,6 +369,9 @@ <h2 id="qunit-userAgent"></h2>
 		<input id="radiocheckbox-1-2" autocomplete="off" type="checkbox" name="radiocheckbox-1" required="required">
 		<input id="radiocheckbox-1-3" autocomplete="off" type="checkbox" name="radiocheckbox-1" required="required">
 	</form>
+	<form id="validatorDestroy" autocomplete="off">
```
jzaefferer: Does this really need a new element? Looks like you could just reuse any other form with a require input.

Westie:
Comment to alert @jzaefferer.

I've had a look around the unit test, bound it to #form instead of a separate form. I'd rather test the event on submit() as that's where I've seen issues in the past (with events not being removed?), however, I've changed it so that it's more like your example.

staabm:
@jzaefferer any progress here?

jzaefferer:
Just landed this! Rebased into a single commit with an updated commit message, also removing the jQuery 1.6.4 file and updating the dependencies entry in `bower.json`.

PR #1394, opened 2015-02-04
throw better errors for bad rule methods

ryleyb:
99% of type errors in this situation are misspelled rules, so be
specific in the throw


jzaefferer:
Thank you for the contribution! Apparently jscs still doesn't like this - could you update this to fix the failure?

PR #1396, opened 2015-02-06
Binding to same unified object as '$'

CoderAjay:
All other validation methods are bound to $ but this one was bound to 'JQuery' , it is causing error when you have two versions of jQuery and want to bind all these methods to one of jQuery instance.


jzaefferer:
Thanks!

PR #1406, opened 2015-02-19
Core: Use !== for null comparison

jimmykane:
This fixes jslinters reporting warning about ```null``` comparison when library is included in frameworks. 

```null``` comparison should be strict.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1072,12 +1072,12 @@ $.extend( $.validator, {
 
 		if ( $.validator.autoCreateRanges ) {
 			// auto-create ranges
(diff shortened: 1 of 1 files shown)


staabm:
this breaks a unit tests. It seems `0` or `false` is also valid in this case.

jimmykane:
@staabm  Correct. Did not see that. Cancel this please

jzaefferer:
`!= null` matches both `null` and `undefined`. jshint accepts that via the `eqnull` option.

jimmykane:
That is true.

PR #1411, opened 2015-02-20
Core: Exclude some keyboard keys to prevent revalidate the field

Arkni:
Avoid revalidate the field when pressing one of the following keys in `onkeyup` method:
```
	Shift           => 16
	Alt             => 18
	Caps lock       => 20
	End             => 35
	Home            => 36
	Left arrow      => 37
	Up arrow        => 38
	Right arrow     => 39
	Down arrow      => 40
	Insert          => 45
	Num lock        => 144
	AltGr key       => 225
```

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -259,7 +259,25 @@ $.extend( $.validator, {
 			}
 		},
 		onkeyup: function( element, event ) {
-			if ( event.which === 9 && this.elementValue( element ) === "" ) {
+			// Avoid revalidate the field when pressing one of the following keys
+			// Shift       => 16
+			// Alt         => 18
+			// Caps lock   => 20
+			// End         => 35
+			// Home        => 36
+			// Left arrow  => 37
+			// Up arrow    => 38
+			// Right arrow => 39
+			// Down arrow  => 40
+			// Insert      => 45
+			// Num lock    => 144
+			// AltGr key   => 225
+			var excludedKeys = [
+				16, 18, 20, 35, 36, 37,
+				38, 39, 40, 45, 144, 225
+			];
+
+			if ( event.which === 9 && this.elementValue( element ) === "" || excludedKeys.indexOf( event.which ) !== -1 ) {
 				return;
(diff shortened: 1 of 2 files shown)


staabm:
Whats the reasoning behind this change?

Arkni:
These keys are used just to navigate, change character's case ...

This change is just a small enhancement to gain in performance and avoiding calling the
`this.element( element )` method every time these keys are pressed.

staabm:
Sounds good to me. To land it requires a unit test which tests at least one of those "special-characters".

@jzaefferer are you ok with this change?

src/core.js
```diff
@@ -259,7 +259,25 @@ $.extend( $.validator, {
 			}
 		},
 		onkeyup: function( element, event ) {
-			if ( event.which === 9 && this.elementValue( element ) === "" ) {
+			// Avoid revalidate the field when pressing one of the following keys
+			// Shift       => 16
```
staabm: What about `ctrl`

Arkni: The  `ctrl` key trigger the `keyup` event when used with another key as combination ( `ctrl + c`, `ctrl + v`, ...)
Do we have to check for it also ?

Arkni:
@staabm 
which file I have to edit in order to add a test suite for this change ?

jzaefferer:
This looks reasonable, should help especially with the remote method.

Since Ctrl by itself doesn't change the input value, it seems good to exclude (add to the list) as well.

As for unit tests: There are two tests in `test/test.js` that set the `onkeyup` option and trigger `keyup` events. Here you'd also need to set the `keyCode` property, like this:

```js
var e = jQuery.Event( "keyup", { keyCode: 16 } );
```

Arkni:
@jzaefferer 

Done.

src/core.js
jzaefferer: This is resolved, Ctrl was added.

jzaefferer:
Nice! Squashed and merged, with a small update of the commit message.

Issue #1414, opened 2015-02-24
Bug with dateITA and daylight saving time in New Zealand


jeredmasters:
I've found another issue relating to date validation.
v1.13.1 - 10/14/2014
This relates to bug #264, where I've also posted this information.

### Problem
We had some customers in New Zealand complaining they couldn't submit the form.
This was reproducible by changing the system timezone on my OSX. 

### Cause
While New Zealand is in daylight savings time their time zone is GMT+13:00.
When javascript processes the "getUTCDate()" it applies -13:00 time change.

The validation function sets the time to 12 noon while parsing, then the -13 hour shift takes the date to 23:00 the previous day.
Then the if statement "xdata.getUTCDate() === gg" fails and says the date is invalid.

### Solution
Our solution was to change the date parser in 
src/additional/dateITA.js:29

```
xdata = new Date(aaaa, mm - 1, gg, 12, 0, 0, 0);
```
to
```
xdata = new Date(Date.UTC(aaaa, mm - 1, gg, 12, 0, 0, 0));
```

I've tested this with several different timezones and it seems to work well.

(Note: I didn't actually change the jquery.validate js, I duplicated the validation method into another js file, gave it a new name, and modified it there)

jzaefferer:
Could you send a pull request for that change? If possible, including unit tests?

jeredmasters:
https://github.com/jzaefferer/jquery-validation/pull/1416

tractorcow:
I can also confirm this bug; Working from NZ the unit tests fail. :)

PR #1416, opened 2015-02-25
Fixed timezone issue for new zealand (GMT+13)

jeredmasters:
Fixed new zealand timezone issue

https://github.com/jzaefferer/jquery-validation/issues/1414#


tractorcow:
I've run this test locally (in the NZ gmt +13 timezone) tested both in and out of daylight saving zones and it fixed the noted issues. Prior to this fix dateITA would fail.

tractorcow:
Is there any way of mocking timezones? http://schickling.me/override-system-time-in-javascript/ is a start. :D

jzaefferer:
A unit test would be nice, but at this rate I figured its better to just merge it. Thanks for the fix and for testing it!

PR #1417, opened 2015-02-26
Add stopPropagation setting

dorner:
When integrating with other JS libraries, such as those that do Ajax requests on form submit, we want to make sure that the form does not actually get submitted when it is invalid. We do this by calling ``stopImmediatePropagation()`` on the event when this setting is true. This stops all other listeners from getting the submit event and therefore the form is not submitted.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -80,6 +80,10 @@ $.extend($.fn, {
 					}
 					return handle();
 				} else {
+				        if ( validator.settings.stopPropagation ) {
+				          // don't bubble submit request
+				          event.stopImmediatePropagation();
+				        }
 					validator.focusInvalid();
 					return false;
 				}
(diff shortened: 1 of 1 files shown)


dholdren:
:+1: I'd really like/need this functionality, and I like that it's optionally set via a flag.

staabm:
I dont think this is something which should be a setting on jquery-validate. you could already achieve the same thing by providing a custom event-handler/submit handler and stop-propagation on the submit event on your own (without requiring code in this lib and adding another setting)

dorner:
Adding external event handlers is tricky because you need to make sure everything gets executed in exactly the right order. It's an invitation for a lot of headaches.

staabm:
I dont think this makes sense. You could also check in your form-submitting logic whether the form has the valid-class.

PR #1423, opened 2015-03-02
Handle event if triggered from inside target, fix #1370

### LINKED ISSUE #1370: Problems with jQuery UI buttons in Webkit browsers due to event.target being SPAN instead of BUTTON
jQuery UI pimps &lt;BUTTON&gt; tag, inserting &lt;SPAN&gt; inside when creating a Button widget. In Firefox and IE, when user clicks on a button (both plain and widget), target.event is always BUTTON itself. In Webkit, however, in case of the widget, target.event is SPAN. Hence, validateDelegate() misbehaves in this case and doesn't run the :submit handler. As the result, formnovalidate attribute and class="cancel" are ignored, and in the submitHandler() function, this.submitButton is not set.

Demo, detailed description and possible solution are here:
http://jsfiddle.net/snobo/Lv0h0yxk/7/

Korri:
Created a fix for issue #1370, we had the same problem, with ou custom buttons having a span inside, I preffered to edit validateDelegate has it should help make it work as intendent in all situations.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -29,12 +29,12 @@ $.extend($.fn, {
 					validator.submitButton = event.target;
 				}
 				// allow suppressing validation by adding a cancel class to the submit button
-				if ( $( event.target ).hasClass( "cancel" ) ) {
+				if ( $( this ).hasClass( "cancel" ) ) {
 					validator.cancelSubmit = true;
 				}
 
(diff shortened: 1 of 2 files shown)


staabm:
Please add a unit test

jzaefferer:
Due to recent changes, switching from custom delegation to using `.on`, this became much easier to fix.

PR #1429, opened 2015-03-06
Reentrancy issue with remote method

ahamid:
While testing a custom variant of the remote method i found that I had to move up this guard otherwise `previous.originalMessage` would get stomped on a reentrant invocation.  To be clear, I haven't confirmed this using the `remote` method directly but I am using all the same boilerplate and just changing my method name (and some details about how the response is handled).
Also I believe `previous.message` should not be getting assigned a resolved value in the case it's a function, otherwise it will always remain that static value instead of dynamically updating on the next error.
I figured it was worth mentioning.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1274,6 +1274,11 @@ $.extend( $.validator, {
 			var previous = this.previousValue( element ),
 				validator, data;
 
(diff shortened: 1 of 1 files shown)


staabm:
Could you post a jsfiddle which shows the problem?

PR #1439, opened 2015-03-23
Regex fix in extension.js

rskm1:
Fix regex in extension function to match PERIOD, rather than "any single character".


jzaefferer:
Thanks!

PR #1440, opened 2015-03-23
Regex fix in accept.js

rskm1:
Fix regex in accept function to match (optional) PERIOD, rather than (optional) "any single character".


jzaefferer:
Thanks!

PR #1441, opened 2015-03-23
Allow "accept" filetype validation to contain a blank among the accepted MIME types

rskm1:
The backstory on this change is that on the Windows platform, if you install WinRAR (or WinZip or 7Zip), those tools will very often screw up the extension-to-mimetype mapping of the OS, so the file object's .type has an EMPTY STRING value when it gets here for validation.

The intuitive thing to do is support EMPTY STRING by just putting a comma prefix on the accept value, such as "accept=',application/zip,application/x-zip-compressed'".  But unfortunately that turns the regex into a mess that will accept ANY type.

This patch handles the case of a blank "accept" type before RegExp has to.

(Also, the RegExp has a period where I think a "\\." should be, but I made that change in a separate pull req since it's unlikely to be controversial, whereas this one may raise some eyebrows and require further discussion).

### DIFF
--- a/src/additional/accept.js
+++ b/src/additional/accept.js
@@ -19,7 +19,7 @@ $.validator.addMethod("accept", function(value, element, param) {
 		if (typeParam.match(/^\|/) || typeParam.match(/\|\|/) || typeParam.match(/\|$/)) {
 			blankTypeAllowed = true;
 			// remove the error-prone blank entry from regex before continuing.
-			typeParam = typeParam.replace(/^\|/, "").replace(/\|\|/g, "|");
+			typeParam = typeParam.replace(/^\|/, "").replace(/\|\|/g, "|").replace(/\|$/, "");
 		}
 
 		// Check if the element has a FileList before checking each file
--- a/src/additional/accept.js
+++ b/src/additional/accept.js
@@ -3,6 +3,7 @@ $.validator.addMethod("accept", function(value, element, param) {
 	// Split mime on commas in case we have multiple types we can accept
 	var typeParam = typeof param === "string" ? param.replace(/\s/g, "").replace(/,/g, "|") : "image/*",
 	optionalValue = this.optional(element),
+	blankTypeAllowed = false,
 	i, file;
 
 	// Element is optional
@@ -15,7 +16,6 @@ $.validator.addMethod("accept", function(value, element, param) {
 		typeParam = typeParam.replace(/\*/g, ".*");
 
 		// Allow checking for BLANK mimetype.
-		var blankTypeAllowed = false;
 		if (typeParam.match(/^\|/) || typeParam.match(/\|\|/)) {
 			blankTypeAllowed = true;
 			// remove the error-prone blank entry from regex before continuing.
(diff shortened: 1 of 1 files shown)


jzaefferer:
Thank you for the contribution. I've merged the other two PRs. This one is certainly more controversial. I don't like that its difficult to even reproduce the issue, so I can't really test this change. Since this is in the "additional" category, I'm okay with landing it anyway.

@staabm any thoughts on this?

src/additional/accept.js
```diff
@@ -14,11 +15,23 @@ $.validator.addMethod("accept", function(value, element, param) {
 		// If we are using a wildcard, make it regex friendly
 		typeParam = typeParam.replace(/\*/g, ".*");
 
+		// Allow checking for BLANK mimetype.
+		if (typeParam.match(/^\|/) || typeParam.match(/\|\|/)) {
```
staabm: I guess this wont work when the EMPTY STRING is the last element in the param?

staabm: Do we also need to handle "space only" comma delimited values like a empty qualifier?

rskm1: Oooo, good catch!  I overlooked the possibility of ending the parm with a comma separator, and that should definitely have the same effect as a double-comma or a comma prefix.

Let me see if I can figure out how to revise a pull req from this web interface...

staabm: Just do another commit at the same branch. We can squash it later when everything looks fine.

Could you also add a few unit tests?

rskm1:
Here's an illustration of why those blank choices need to be handled separately by this code and NOT by RegExp.  Basically, the RegExp library that javascript uses has a known bug in it.

Javascript console excerpt:
> function m(p,t) { console.log(p.match(new RegExp( ".?(" + t + ")$", "i"))) }
> m('gif', 'jpg|jpeg')
null
> m('gif', 'jpg||jpeg')
["f", "", index: 2, input: "gif"]

As you can see, RegExp "finds" a match where there really ISN'T a match, just because of the double-| (or blank prefix | or blank suffix |) in there.

src/additional/accept.js
rskm1: Done!

As for unit tests, I poked around in the tests directory and couldn't find any existing tests of the "accept" method. There are lots of tests for the "extension" method in /test/methods.js, but... I'm not super familiar with this codebase but I thought "accept" was intended to supercede "extension"?

staabm:
I am fine with the change, only thing missing is a test.

jzaefferer:
There are tests for the `accept` method in #1373, which fixes another problem with the method. @rskm1 can I interest you in checkout out that PR and merging it with your changes?

rskm1:
Oh, very nice!  #1373 looks good to me, but I'm not sure how to merge someone else's branch into my own using github's web interface.  If you go ahead and merge #1373 into master, this PR will have merge conflicts that I'm pretty sure I can figure out how to clean up, and add more "accept" test cases to.

jzaefferer:
Well, you can't run the tests from the GitHub web interface either (right?). Making a local clone and editing there shouldn't be a big barrier. Let me know if you need help with that.

jzaefferer:
@rskm1 are you still interested in this? Your help is definitely very welcome.

staabm:

PR #1444, opened 2015-03-28
Added composer.json

royopa:


jzaefferer:
Why?

royopa:
@jzaefferer Composer is a dependency manager for PHP language. Adding the composer,json enable install the jquery-validation with Composer.
More info here: https://getcomposer.org/ and here: https://packagist.org

jzaefferer:
Sorry for being uncooperative, but I don't want to maintain more files like this. The same information is available via the existing `package.json`.

PR #1453, opened 2015-04-09
messages_ge.js added Georgian Language

tatocaster:
added Georgian (ქართული) language translations for messages


jzaefferer:
Thanks!

PR #1461, opened 2015-04-16
Additional validation `time` : the hour part can to be only one digit

Fridus:


jzaefferer:
Thanks! I've added an assertion to the unit tests and updated the commit message.

PR #1470, opened 2015-04-27
Cast NaN value attributes to undefined for Opera Mini

marek-saji:
Opera Mini failed to validate fields with no `minlength` attribute specified producing error messages such as:

> Please enter at least NaN characters.

Setting `NaN` values to `undefined` fixes the problem.


jzaefferer:
Thanks!

PR #1474, opened 2015-04-30
Descendants of a disabled fieldset are not ignored

weitzhandler:
Hi,

Change in [line 553](https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L553):

Substituted `[disabled]` with `:disabled`, this is crucial, because `[disabled]` doesn't affect fields that are under a disabled `fieldset`.

Currently the embedded input bellow is being validated, because it doesn't fulfil `[disabled]`.
It does however return `true` for `$('#disabled').is(':disabled')`, hence the reason of my change.

    <fieldset disabled="disabled">
      <input id="disabled" type="text" />
    </fieldset>

And as a workaround, until this issue is fixed, `ignore: ":hidden, fieldset[disabled] *"` can be used.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -550,7 +550,7 @@ $.extend( $.validator, {
 			// select all valid inputs inside the form (no submit or reset buttons)
 			return $( this.currentForm )
 			.find( "input, select, textarea" )
-			.not( ":submit, :reset, :image, [disabled]" )
+			.not( ":submit, :reset, :image, :disabled" )
 			.not( this.settings.ignore )
 			.filter( function() {
 				if ( !this.name && validator.settings.debug && window.console ) {


jzaefferer:
Thanks! I've replaced the commit message, but otherwise landed your change as-is.

weitzhandler:
@jzaefferer 
>I've replaced the commit message, but otherwise landed your change as-is.

Didn't really understand what you mean by that, is the bug gonna fixed or not?

staabm:
@weitzhandler it landed on master (and therefore is fixed in the next release)

see https://github.com/jzaefferer/jquery-validation/commit/f37c2d8131ca0e3c2af0093f6fd9d2c40c282663

would be great if you could also add a unit test which reproduces the initial issue (as a separate PR)

weitzhandler:
I will add unit tests.
I was just wondering why it says
>Closed with unmerged commits
This pull request is closed, but the weitzhandler:patch-1 branch has unmerged commits.

jzaefferer:
I rebased your commits on top of other changes in master. That avoids merge commits, while the commit itself still references this PR through the `Closes ...` line.

PR #1475, opened 2015-05-01
Fix Finnish translation

Pietu1998:
The current Finnish translations are partially wrong and partially clumsy. This change should improve them. Finnish is really a hard language to equivalently and clearly translate some strings to.


jzaefferer:
Thanks!

PR #1482, opened 2015-05-12
Builds: Adds prebuilt files to repository

### LINKED ISSUE #1481: Request: add prebuilt jquery.validate.js to npm repository
As a gulp user who is hoping to use `npm` without `bower` (as an increasing number of people are doing), I would like to continue using this awesome jQuery plugin to validate my forms so that I don't have to rewrite a bunch of code or add extra steps to my build process. 

Criteria:
- A usable version of jquery.validate.js exists and is ready to use directly from the npm module without having to do an additional build step

It would be super awesome to have this change implemented. Let me know if I can help! Thanks!

sbussard:
To allow use of npm as a front end package manager

Closes #1481

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
-dist
 node_modules
--- a/dist/additional-methods.js
+++ b/dist/additional-methods.js
@@ -1,0 +1,998 @@
+/*!
+ * jQuery Validation Plugin v1.13.2-pre
+ *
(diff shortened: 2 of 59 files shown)


sbussard:
Added this to package.json instead `"jquery-validation": "git+ssh://git@github.com:jzaefferer/jquery-validation.git#1.13.1"`

PR #1484, opened 2015-05-14
Add a copy of the MIT license

Calinou:
It is good practice to add a copy of the license to the repository.


jzaefferer:
Thanks! I've updated it to remove the year, since that is unnecessary and annoying to update each year (following the lead of the jQuery Foundation here, where the same change was made).

PR #1489, opened 2015-05-26
Package: Adding build script to npm package.json

chuym:
Adding build script to package.json, this way when the package is imported through npm, it is ready to use.


jzaefferer:
Thanks for the contribution. The problem being addressed here was already fixed by adding the `files` property. The next release will have the necessary files published to npm.

PR #1490, opened 2015-05-27
cancel and formnovalidate attributes do not work for buttons with inner elements

javcz:
A common scenario <button type="submit" class="cancel"><span class="maybesymbolfont">x</span><button> will not work as expected, validation does not get cancelled. Proposed fix searches for the appropriate target when the inner element is clicked

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -24,17 +24,24 @@ $.extend($.fn, {
 
 		if ( validator.settings.onsubmit ) {
 
-			this.validateDelegate( ":submit", "click", function( event ) {
+			this.validateDelegate( ":submit, :submit *", "click", function( event ) {
 				if ( validator.settings.submitHandler ) {
 					validator.submitButton = event.target;
 				}
+				
+				// locate the actual "action" element
+				var actionTarget = $(event.target);
+				if(!actionTarget.is(":submit")) {
+				    actionTarget = actionTarget.closest(":submit");
+				}
+				
 				// allow suppressing validation by adding a cancel class to the submit button
-				if ( $( event.target ).hasClass( "cancel" ) ) {
+				if ( actionTarget.hasClass( "cancel" ) ) {
 					validator.cancelSubmit = true;
 				}
 
 				// allow suppressing validation by adding the html5 formnovalidate attribute to the submit button
-				if ( $( event.target ).attr( "formnovalidate" ) !== undefined ) {
+				if ( actionTarget.attr( "formnovalidate" ) !== undefined ) {
 					validator.cancelSubmit = true;
 				}
 			});
--- a/src/core.js
+++ b/src/core.js
@@ -28,13 +28,13 @@ $.extend($.fn, {
 				if ( validator.settings.submitHandler ) {
 					validator.submitButton = event.target;
 				}
-				
+
 				// locate the actual "action" element
 				var actionTarget = $(event.target);
 				if ( !actionTarget.is( ":submit" ) ) {
 				    actionTarget = actionTarget.closest(":submit");
 				}
-				
+
 				// allow suppressing validation by adding a cancel class to the submit button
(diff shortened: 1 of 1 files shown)


javcz:
Oops, forgot to add in the original comment. I only was able to reproduce this behavior in the Chrome browser. Only Chrome has in the event.target property set to the inner button element when listened for the click on form. Other tested browsers (Firefox, IE11) have the event.target set to the button itself and the problem does not show there.

javcz:
jsbin demonstrating the bug: http://output.jsbin.com/pagutolovo/10 

The second button should not trigger the validation, however if you click the blue text on the button (make sure you hit the glyphs, not the button face), the validation will run. Chrome only (from what I have tested)

javcz:
Sorry guys, I found too late this is a duplicate to https://github.com/jzaefferer/jquery-validation/issues/1370, just a slightly different solution of the problem

src/core.js
```diff
@@ -24,17 +24,24 @@ $.extend($.fn, {
...
 		if ( validator.settings.onsubmit ) {
 
-			this.validateDelegate( ":submit", "click", function( event ) {
+			this.validateDelegate( ":submit, :submit *", "click", function( event ) {
 				if ( validator.settings.submitHandler ) {
 					validator.submitButton = event.target;
 				}
+
+				// locate the actual "action" element
+				var actionTarget = $(event.target);
```
staabm: couldn't we just check wheter `event.target` is of type input, and if not, `find` for the input starting from `event.target`?

staabm:
please integrate the unit test of https://github.com/jzaefferer/jquery-validation/pull/1392/files to make sure you are actually fixing the issue and we wont regress.

src/core.js
javcz: I think this is exactly what the code does. With the extension in must not limit the test to input only and needs to work for `<button type=”submit”>` also. Basically exactly what :submit selector test does. This solution (=use the :submit selector) seems cleaner and more readable to me. Moreover at this point there is no need to worry about the performance.
As for the performance I did not measure the impact of the validateDelegate compound selector, that one I would suspect first.
Just my opinion, if I’m missing anything, please, let me know.

staabm: Right, but I want to achieve it in a different way as implemented currently. Instead of binding to `:submit *` we should do a find only when event.target doesnt work.

This would help because only chrome needs to pay for this edge case, not all browsers

staabm: Hmmm after re-reading it looks like everything is in place.

Please just add the unit test

javcz: OK, understand, that makes sense but it implies the other solution described in the issue-1370 … getting rid of the validateDelegate, which now filters out the click events coming from the inner elements. That solution would be more efficient however (and a bigger change in the current code)

javcz: oops, soory, I have missed your previous reply ... ignore the post above ....

javcz:
Hmm, sorry, I’m lost now. The PR-1392 seems to me to primary focus a quite different topic (clean removing of the validator from DOM) and it “by the way” replaces the `validateDelegate` by `on` event binding, which I believe solves the Webkit problem in a natural way (as the click event will bubble form the inner element and the handler will be invoked when reaching the parent :submit level – something what will not happen in the current implementation). 
Just this solution is a bigger change to the current code and I understood that is the reason why you prefer to keep the `validateDelegate`. Once you plan to pull PR-1392 the change I have proposed is not needed anymore. Where exactly would you like me to add the unit test method? 
I’m a total newbie in contributing to open projects, sorry for silly questions …

staabm:
Lets hold on a bit and see what jzaeffer responds in the alternative PR. Thanks for your contribution.

jzaefferer:
Due to recent changes, switching from custom delegation to using `.on`, this became much easier to fix.

Issue #1491, opened 2015-05-30
Rule 'remote' complete response data basically unusable


rain01:
The issue is that right now only valid response for remote is true and false.
That leaves [remote 'complete' function](http://stackoverflow.com/a/16214487/1537657) almost useless. 
For example if I want to do this, it fails:
```javascript
$('#my_form').validate({
    onkeyup: false,
    rules: {
        username: {
            remote: {
                url: '/validate/',
                complete: function(resp, status){
                    if (status == 'success') {
                        var data = $.parseJSON(resp.responseText);
                        if (typeof data === 'object' && 'avatar' in data)  {
                            $('#avatar').attr('src', data['avatar']);
                        }
                        else {
                            $('#avatar').attr('src', 'default.png');
                        }
                    }
                }
            }
        }
    },
    messages: {
        username: {
            remote: 'This user does not exist.'
        }
    }
});
```
It fails because I can only send true or false from server.

I would love to see it more useful if I could send more data from server for the 'complete' function to use.
For example sending JSON from server:
```javascript
{
    valid: true,
    whatever_else_I_want: 'here is my data'
}
```
Where valid acts the way it acts right now validating if form is valid or not and with rest of the data you can do what ever you want in the 'complete' function making it much more useful.

PS: I could not find any documentation about remote 'complete' function so I posted stackoverflow link.

jzaefferer:
Have you tried using the `dataFilter` option from [$.ajax](http://api.jquery.com/jQuery.ajax/)?

> A function to be used to handle the raw response data of XMLHttpRequest. This is a pre-filtering function to sanitize the response. You should return the sanitized data. The function accepts two arguments: The raw data returned from the server and the 'dataType' parameter.

rain01:
I have but I couldn't figure it out how to use it for the life of me. I returned the response but couldn't figure out where it was sent since it never got to complete... complete still gave the same response.

jzaefferer:
Can you share what you came up with?

rain01:
I tried all kinds of different stuff but since it didn't work I deleted it. I'll add how I thought it should work.

```
username: {
    remote: {
        url: '/validate/',
        dataFilter: function(data, type) {
            // data = 'does not matter what I give from server'
            return true; // sending to complete
        },
        complete: function(resp, status) {
            // resp.responseText = 'does not matter what I give from server'... but I expect TRUE (from dataFilter)
            console.log(resp.responseText);
        }
    }
}
```

[...]

PR #1493, opened 2015-05-30
Feature: Make remote complete function more useful

rain01:
Added an option to remote rule to send more data from server than just valid status and error message.

This feature is backwards compatible, meaning everything will still work without changing your code but now you could make it more advanced if your server response is this...

Server json response:
```javascript
{
    valid: true,
    whatever: 'my random data'
}
```

Feature request #1491

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1299,6 +1299,9 @@ $.extend( $.validator, {
 				data: data,
 				context: validator.currentForm,
 				success: function( response ) {
+					if (typeof response === 'object' && 'valid' in response) {
+						response = response['valid'] === true || response['valid'] === "true";
+					}
 					var valid = response === true || response === "true",
 						errors, message, submitted;
 


staabm:
This doesn't look like the way to go. We should not dictate the strutcture of the JSON required.

Maybe we should add a callback instead which can decide - using `return true/false` - whether the returned response was "valid"?

jzaefferer:
I agree that this format is too specific. A callback would be more flexible. There's also the relate PR #1353. I don't have an answer yet.

rain01:
@staabm JSON is not required but optional... you can still return true\false or str but when you return JSON you can add more data to it.

Adding a callback would be a different feature, probably wouldn't satisfy #1491 and it would be much more complicated.

staabm:
> Adding a callback would be a different feature, probably wouldn't satisfy #1491 and it would be much more complicated.

why wouldn't this satisfiy the ticket?
you would have a callback for the response (implemented in your app, instead of the lib) which checks the JSON response and returns `true/false` whenever the JSON contains a `valid: true/false` property ?

rain01:
> your logic atm dictates how the json is interpreted.

That's how JSON is... have to interpret it some way... no?

> One issue with your approach is e.g. when someone returns a object with a property named valid the interpretation would be completly different, after your change would land in the library. 

Right now if you return an object, it would do nothing doesn't matter if there is 'valid' or not in there... that's why I added the new feature where you could return an object.

> Additionally this renders to be a BC break.

How would it be BC break? You can still do everything you do now returning bool or str and it would work the same way... only when you return JSON you have new options as oppose to not doing anything.

>why wouldn't this satisfiy the ticket?
you would have a callback for the response (implemented in your app, instead of the lib) which checks the JSON response and returns true/false whenever the JSON contains a valid: true/false property?

Maybe I don't understand how you would do that but isn't remote's "complete" already a callback? Where would you put that callback in #1491 situation?

If the callback is the answer then can someone do it or show how it's done?

jzaefferer:
This isn't a good enough solution. I've also commented on the underlying issue.

PR #1497, opened 2015-06-05
CONTRIBUTING: Added a hint on how to run the testsuite

staabm:
using php

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,8 @@ If you've wrote custom methods that you'd like to contribute to additional-metho
 
 ## Unit Tests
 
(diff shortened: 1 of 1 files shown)


jzaefferer:
PHP isn't a requirement anymore, since all remote requests are mocked with mockjax. Could you update this to just remove the need for a local server?

staabm:
Using this webserver php local webserver it was very easy for me to run the tests (so no grunt required)

Nevertheless you prefer the way running it using grunt and without a webserver, right?

jzaefferer:
No, I meant that you can just run `test/index.html` from the filesystem in your browser, no webserver nor grunt required.

staabm:
Using the `file:///`protocol?

jzaefferer:
Yep

staabm:
@jzaefferer here we go

jzaefferer:
Thanks!

PR #1501, opened 2015-06-17
added license info to package.json

staabm:
This fixes a warning while `npm install` that the package does not provide license information

>npm WARN package.json jquery-validation@1.13.2-pre No license field.


jzaefferer:
Thanks!

PR #1509, opened 2015-06-26
Core: Added validAsync method to core

danieloprado:


src/core.js
```diff
@@ -1187,7 +1212,7 @@ $.extend( $.validator, {
 
 		// http://jqueryvalidation.org/number-method/
 		number: function( value, element ) {
-		    return this.optional( element ) || /^(?:-?\d+|-?\d{1,3}(?:,\d{3})+)?(?:\.\d+)?$/.test( value );
+			return this.optional( element ) || /^(?:-?\d+|-?\d{1,3}(?:,\d{3})+)?(?:\.\d+)?$/.test( value );
```
jzaefferer: I've fixed this in master.

src/core.js
```diff
@@ -106,10 +106,35 @@ $.extend($.fn, {
...
+	validAsync: function(callback) {
+		var validator, interval, valid = this.valid();
+
+		if ( $( this[ 0 ] ).is( "form" ) ) {
+			validator = this.validate();
+		} else {
+			validator = $( this[ 0 ].form ).validate();
+		}
+
+		if ( typeof(callback) !== "function" ) {
```
jzaefferer: Adding a separate method seems okay, but it doesn't make sense to have an optional callback then. Let's drop this check and return.

jzaefferer:
Thank you for the contribution. This is a good start. Could you look into adding a test for this new method?

danieloprado:
Ok, I removed the callback verification and I'll added a test as soon as possible
Thanks for the reply

jzaefferer:
Just comment here again when you've updated the PR, or if you need help with the test.

PR #1510, opened 2015-06-28
Allow for protocol-relative URLs

yeslad:


jzaefferer:
Thank you for the contribution. Could you add a few assertions to the url unit test to cover this change?

yeslad:
Done.

jzaefferer:
Thanks!

Issue #1111, opened 2014-05-10
File Size


arunsasi:
How to check file size?

twolfson:
If anyone is still interested, I have created #1512 in order to create a maximum file size plugin

Issue #450, opened 2012-06-28
Check file size for upload in supported browsers using HTML5 File API


niftylettuce:
This is a mix between jquery-validation and Modernizr, although we can port the Modernizr part out so Modernizr is not a dependency.

I'm wondering if you are open to having me add it to additional-plugins?  It is a great feature :)

```js
  // Validate file input on change
  jQuery.validator.addMethod("filesize", function(value, element, param) {
    // This requires Modernizr filereader community add-on:
    // <https://github.com/Modernizr/Modernizr/blob/master/feature-detects/file-api.js>
    var valid = true;
    if (Modernizr.filereader) {
      // Check if file size is valid
      // (e.g. value = "10485760" = 10mb)
      for(var i=0;i<element.files.length;i+=1) {
        if (element.files[i].size > param) valid = false;
      }
    }
    return this.optional(element) || valid;
  }, "File exceeds maximum upload size.");
```

jzaefferer:
I'm sorry for the lack of activity on this issue. Instead of leaving it open any longer, I decided to close old issues without trying to address them, to longer give the false impression that it will get addressed eventually.

To the reporter (or anyone else interested in this issue): If you're affected by the same issue, consider opening a new issue, with a testpage that shows the issue. Even better, try to fix the issue yourself, and improve the project by sending a pull request. This may seem daunting at first, but you'll likely learn some useful skills that you can apply elsewhere as well. And you can help keep this project alive. [We've documented how to do these things, too](https://github.com/jzaefferer/jquery-validation/blob/master/CONTRIBUTING.md). A patch is worth a thousand issues!

twolfson:
If anyone is still interested, I have created #1512 in order to create a maximum file size plugin

PR #1512, opened 2015-07-08
Additionals: Added maxsize plugin

twolfson:
We recently created a maxsize plugin to detect that a file is too large before submission. Based off of #450 and #1111, it looks like other people would like it as well. In this PR:

- Added maxsize plugin into `src/additional`

**Missing:**

We didn't add any tests as there are no tests for the only other `FileList` based plugin (`accept.js`). I am assuming that the reason for this is we cannot programmatically tell a browser where to get our file uploads from.

### DIFF
--- a/src/additional/maxsize.js
+++ b/src/additional/maxsize.js
@@ -1,0 +1,30 @@
+// Require a file at/under a given size
+// Usage: `{rules: {file: {maxsize: 5 * 1024 * 1024 /* 5MB */}}}`
+$.validator.addMethod("maxsize", function(value, element, param) {
+  // If the value is optional, return it
+  var maxSize = param,
+      optionalValue = this.optional(element),
+      i, len, file;
+  if (optionalValue) {
+    return optionalValue;
+  }
+
+  // If the element is a file, then verify the size is at most what we expect
+  if ($(element).attr("type") === "file") {
+    // If we are in a browser that supports FileList, then check the files
+    if (element.files && element.files.length) {
+      i = 0;
+      len = element.files.length;
+      for (; i < len; i++) {
+        file = element.files[i];
+        // `file.size` is size in bytes
+        if (file.size > maxSize) {
+          return false;
+        }
+      }
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


jzaefferer:
Thank you for the contribution. This looks like a good start, but very much needs tests. There is an open PR that added tests for the accept method (#1373) and one more that further extends the method (#1441). Both need some rebasing or other fixes to get merged. Could I interest you in that? If so, start with the accept PRs (rebase and whatever else is needed), then rebase this PR on top and share the testing infrastructure (the `acceptFileDummyInput` function).

twolfson:
imo, per-computer filetype validation isn't too trustworthy as it can change a lot from computer to computer (e.g. we have seen `.pdf` with the type of `application/download`, `www/unknown`, `application/x-pdf`). As a result, I would prefer to stay away from #1441 and its edge cases.

Is there anything that needs to be done in #1373 aside from a squash/rebase?

jzaefferer:
No, I think that's all there is to it.

twolfson:
Is there any reason why I should do it over you? I am not quite sure I follow the problem =/ The attribution should still be preserved -- here is an example in `jscs`: https://github.com/jscs-dev/node-jscs/commit/18b4ffce821f03ab04b7df97949ddacedae315a1

twolfson:
If you are unfamiliar with how to grab commits from a PR outside of GitHub, here is a gist describing the details:

https://gist.github.com/piscisaureus/3342247

jzaefferer:
The problem are the 7 commits on that PR, of which at least three produce conflicts during a rebase. I figured you could help resolve those, since you'd likely end up integrating some of those changes into your own PR anyway.

twolfson:
Taking a shot at getting #1373 back on its feet

kpduncan:
Please look at #6834.

twolfson:
This PR was opened over a year ago, I no longer work at the organization I started it from, and I no longer have time to contribute to this PR. Closing this PR -- someone can pull over the changes into another PR if they want them

PR #1515, opened 2015-07-09
Speedup travis builds

staabm:
Activate the faster travis infrastructure and activate caching for npm dependencies between builds.

Speedup does not necessarily mean per build time, but also shorter waiting times in build queues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: node_js
 node_js:
   - 0.10
+
+cache:
+  - npm
+sudo: false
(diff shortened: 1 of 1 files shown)


jzaefferer:
Thanks! I just updated the commit message. And while we're here: You're welcome to push branches that are likely long-lived to this repo, otherwise you can stick with creating branches on your repo. Not that it matters much either way.

staabm:
Thanks, I try to remember it :D

PR #1518, opened 2015-07-10
Added support for custom email patterns, provided via options.emailRegex.

staabm:
Uses the current email regex per default to keep BC.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -234,6 +234,7 @@ $.extend( $.validator, {
 		onsubmit: true,
 		ignore: ":hidden",
 		ignoreTitle: false,
+		emailRegex: /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/,
 		onfocusin: function( element ) {
 			this.lastActive = element;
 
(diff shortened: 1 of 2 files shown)


jzaefferer:
I'm okay with making the email validation configurable, but I really want to avoid adding more options. Partly because there are too many already, partly because this would tie in the method even more into the core.

To take a step back, why do you need this as an option? Could you override the email method as needed? Or use a custom email method when you need to the default on the same page?

staabm:
> To take a step back, why do you need this as an option? Could you override the email method as needed? Or use a custom email method when you need to the default on the same page?

I could of course use a separate email validation method. IMO email validation will only be different in the email format and per definition will not have other differences, so I decided to go this road.

Per se, I would love to reduce the jquery-validation core to less lines and have a built which only contains the rules I really need for a certain app.

staabm:
As you already know, email pattern issues are the most reported errors in this tracker, therefore solving the issue using a config param is a must have for a lot of our users. (I am not saying it has to be a option passed to the validator)

staabm:
@jzaefferer what do you think, how to proceed?

staabm:
I just tried the other way you proposed.

In case we promote the `$.validator.methods.<myRule>` properties to be part of the public api, I could override the native-email method with my custom regex without providing another option.

@jzaefferer wdyt?

jzaefferer:
That looks good to me.

staabm:
will open a doc PR then, thanks.

staabm:
here we go
https://github.com/jzaefferer/jquery-validation/pull/1530
https://github.com/jzaefferer/validation-content/pull/14

PR #1519, opened 2015-07-10
README: use svg badges, for retina friendlyness

staabm:


jzaefferer:
Thanks!

staabm:
FYI I also checked for the pledgie badge, but it seems they dont support SVG badges atm.

Issue #1000, opened 2014-01-31
Date Validation For February Month


anksce:
Hi,

If user enters '02-30-2014' then it doesn't show as invalidate date.
Instead of that it takes '03-02-2014'.

We can do something like this
http://www.jquerybyexample.net/2011/12/validate-date-using-jquery.html

Issue #939, opened 2013-11-14
Invalid date validates


bean5:
According to http://jqueryvalidation.org/contribute/ which uses jquery-validation, this date will validate: 09/31/2013. However, there is no Sept. 31. I don't know how to test whether http://jqueryvalidation.org/contribute/ is using old code or not.

PR #1528, opened 2015-07-18
Core: Add sanity checks to dateISO

RobJohnston:
Added checks for leap years and for months that only have 30 days.

I've re-used the regex for dateISO, but I was talking to @nschonni  and from what we can see, the slashes are not part of the [ISO 8601 standard](https://en.wikipedia.org/wiki/ISO_8601) (but the canonical source is paywalled).  It's either a dash or nothing, so maybe the regex should be `/^(\d{4})(?:\-)*(0?[1-9]|1[012])(?:\-)*(0?[1-9]|[12][0-9]|3[01])$/`.


RobJohnston:
Relates to https://github.com/jzaefferer/jquery-validation/issues/939 and https://github.com/jzaefferer/jquery-validation/issues/1000.

src/core.js
```diff
@@ -1187,7 +1187,33 @@ $.extend( $.validator, {
...
+					check = false;
+				} else if (mm === 2) {
+					isLeap = (yyyy % 4 === 0 && (yyyy % 100 !== 0 || yyyy % 400 === 0));
+
+					if (dd > 29 || (dd === 29 && !isLeap)) {
+						check = false;
+					}
+				}
+			}
+			return this.optional(element) || check;
```
staabm: Please move the optional check in a short-circuit to be consistent with other rules which do the heavy checks only when necessary

RobJohnston:
@staabm : Like so?  I was being consistent with dateITA originally.

jzaefferer:
Thanks for the contribution! While we're at it, let's drop the slashes and update the tests accordingly. I have no idea why they're there. W3C doesn't list anything with slashes either: http://www.w3.org/TR/NOTE-datetime

RobJohnston:
Just drop the slashes or also allow YYYYMMDD in addition to YYYY-MM-DD?

jzaefferer:
Just disallow slashes, only allows dashes as separator.

Arkni:
Does that make sense if this validator accept years between **`0000`** and **`09999`** ?
because _**MySQL 5.x**_ accept dates between **`1000-01-01`** and **`9999-12-31`**.

RobJohnston:
[True for MySQL] (https://dev.mysql.com/doc/refman/5.0/en/datetime.html), but different databases have different ranges (and sometimes dates are not stored in databases):

A quick search:

* Postgres accepts dates between [4713 BC and 5874897 AD] (http://www.postgresql.org/docs/9.4/static/datatype-datetime.html).

* Oracle seems to work between [4712-01-01 BC and 9999-12-31] (http://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#i1847).

* SQL Server accepts dates between [0001-01-01 and 9999-12-31] (https://msdn.microsoft.com/en-us/library/bb677335.aspx).

I would think that the [min method] (http://jqueryvalidation.org/min-method/) and [max method] (http://jqueryvalidation.org/max-method/) would be the places to specify acceptable ranges.

staabm:
Not sure we can handle as invalid which were valid before and vice versa. I would merge such a change only with a new major version

RobJohnston:
But was February 31st ever really valid?

staabm:
at least `1990/06/06`, `1990-6-6` and `1990/6/6` were valid before..

PR #1533, opened 2015-07-22
fix #1532

### LINKED ISSUE #1532: min/max rules are applied to text fields implicitly
I've just met a weird problem. I've got something like this:

```html
<input class="infoTableInput2" name="owner_name" type="text" id="owner_name" value="">
```

and the rule:

```javascript
$('#store_form').validate({
    rules : {
        owner_name: {
            required: true
        }
    },
    messages : {
        owner_name: {
            required: '{$lang.owner_name_required}'
        }
    }
});
```

every time I tried to enter something to the field, an error message pops up `Please enter a value greater than or equal to 0.`. But as you can see above, I've not added any extra rule to the element.
After debugging a bit, I've found that it's because the validator loops through all available methods in the normalizer. And the line 981 will always be true while evaluating the min/max method:

```javascript
if ( /min|max/.test( method ) && ( type === null || /number|range|text/.test( type ) ) )
```

eidng8:
#1532
I think value of methods are not supposed to be empty, so this extra check would do the trick.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -978,7 +978,7 @@ $.extend( $.validator, {
 
 		// convert the value to a number for number inputs, and for text for backwards compability
 		// allows type="date" and others to be compared as strings
-		if ( /min|max/.test( method ) && ( type === null || /number|range|text/.test( type ) ) ) {
+		if ( value !== "" && /min|max/.test( method ) && ( type === null || /number|range|text/.test( type ) ) ) {
 			value = Number( value );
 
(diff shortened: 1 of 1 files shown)


staabm:
@eidng8 any chance to add a unit test to this?

Arkni:
@staabm 

I think a fiddle demonstrating the problem is required here.

staabm:
Closing because of missing response.

PR #1534, opened 2015-07-22
Core: Allow 'false' as value when overriding rules

0x000000:
This PR enables ability to deny or override rules via declarative HTML attributes (as it done for rules defined in `.validate` method).

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1059,7 +1059,7 @@ $.extend( $.validator, {
 		// handle dependency check
 		$.each( rules, function( prop, val ) {
 			// ignore rule when param is explicitly false, eg. required:false
-			if ( val === false ) {
+			if ( val === false || val === "false" ) {
 				delete rules[ prop ];
 				return;
 			}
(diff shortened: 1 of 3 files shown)


staabm:
note sure this is something which makes sense to be configurable using markup.

could you describe your use-case?

0x000000:
@staabm Sure! There is a popular lib for Ruby on Rails: https://github.com/plataformatec/simple_form , which creates selects (only for dates) with predefined `class="date"` attribute. So select will look like `<select class="date" />, nothing criminal for me. By some unknown reason it breaks validation of this plugin for selects in Firefox on Windows (Firefox on Mac works as expected) and we have such selects displayed as "invalid" even if valid option was selected by user. So easy fix is to disable data validation for this inputs in way like `<select class="date" date="false" />`.

staabm:
Could you build a small jsfiddle which reproduces the issue?

staabm:
Programmatically based solution should always override declarative approaches, because those are considered more specific and dynamic. Therefore I will decline this feature.

Please feel free to discuss further and describe your case. I am willing to re-open the issue in case we have good reasons to do so.

PR #1547, opened 2015-08-12
case-insensitive check for BIC

netzkind:
I found no appearant reason why a BIC has to be all-upper, so allowing customers to enter it lowercase or mixed should be ok.
Furthermore, the IBAN-validation also is case-insensitive.

### DIFF
--- a/src/additional/bic.js
+++ b/src/additional/bic.js
@@ -12,5 +12,5 @@
  * - Last 3 characters - branch code, optional (shall not start with 'X' except in case of 'XXX' for primary office) (letters and digits)
  */
 $.validator.addMethod("bic", function(value, element) {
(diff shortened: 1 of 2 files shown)


staabm:
For consistency with IBAN I agree. I think we should add in the comments of both validations that the validation is handled in a case-insensitive manner and that it is up to the user to normalize the input (e.g. convert everything into uppercase before persisting the user input).

netzkind:
Sounds good to me. Will you need an amended pull request/patch?

staabm:
please add the requested change to this PR.

netzkind:
OK, comments are amended.

staabm:
Thanks, your changes look good to me.

staabm:
@netzkind could your squash those commits?

staabm:
@netzkind please squash your commits and prefix the commit message with `Additionals:`

I will merge after this small changes.

staabm:
@netzkind are you willing to apply the above mentioned changes?

PR #1555, opened 2015-08-19
turkish identification number

cerkiner:
Additional method that returns true, if the value is a well formatted
turkish identification number (T.C. Kimlik No).


staabm:
Could you please give us some context what a turkish id actually is?
Is its format defined as a standard?

Could you add some tests?

cerkiner:
sorry for my silliness, i forgot to mention it.
Here is some trustworthy info on [wikipedia](https://www.wikiwand.com/en/Turkish_Identification_Number)

It is an identification method for citizens of turkey. every citizen has one. and yes there is a way to check if it is valid because last two digits are for validation. use can see and compare with my code to understand.

having it here will make sense for the ones looking for it.

staabm:
As we dont yet support citizen numbers for other countries I am not willing to merge this.

our long term goal is to get all these non-core validators out of this repos, therefore I accept only variants of already contained plugins.

PR #1557, opened 2015-08-25
add predict for remote validation response

zdjray:
Sometimes we need to custom the remote validation json in response.
Here we add predict for this situation.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1317,7 +1317,7 @@ $.extend( $.validator, {
 				data: data,
 				context: validator.currentForm,
 				success: function( response ) {
-					var valid = response === true || response === "true",
+					var valid = response === true || response === "true" || (param.predict && param.predict.call(this, response)),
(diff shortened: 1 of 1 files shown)


staabm:
whats your use case? what is this code supposed to do?

zdjray:
the situation is:
the remote REST API has its own response format,just like this:
```javascript
{
    result: 0,
    msg: 'OK'
}
```
or this:
```javascript
{
    result: -1,
    msg: 'some error occurs'
}
```
when result is 0,it means the validation is correct.

staabm:
wouldn't jquerys `dataType` or `dataFilter` option of `jQuery.ajax`[1] do the job for you?

[1] http://api.jquery.com/jQuery.ajax/

zdjray:
thanks for the perfect solution,that will work for me
it's a great idea:D

PR #1558, opened 2015-08-26
Create phoneBR.js

edakagawa:
Validates brazilian phone numbers (without international prefix).

Valida telefones brasileiros, fixos ou celulares, com ou sem DDD.
EX: 11 91234-1234 OU (011) 1234-1234

### DIFF
--- a/src/additional/phoneBR.js
+++ b/src/additional/phoneBR.js
@@ -2,4 +2,4 @@ $.validator.addMethod("phoneBR", function(phone_number, element) {
 	phone_number = phone_number.replace(/\s+/g, "");
 	return this.optional(element) || phone_number.length > 7 &&
 		phone_number.match(/^(\(?[0-9]{2,3}\)?)? ?([0-9]{4,5})-?([0-9]{4})$/gm);
-}, "Digite um numero de telefone");
+}, "Please specify a valid phone number");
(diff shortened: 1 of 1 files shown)


src/additional/phoneBR.js
```diff
@@ -0,0 +1,5 @@
+$.validator.addMethod("phoneBR", function(phone_number, element) {
+	phone_number = phone_number.replace(/\s+/g, "");
+	return this.optional(element) || phone_number.length > 7 &&
+		phone_number.match(/^(\(?[0-9]{2,3}\)?)? ?([0-9]{4,5})-?([0-9]{4})$/gm);
+}, "Digite um numero de telefone");
```
staabm: The default impl should contain a english error message

staabm:
please add a few tests into https://github.com/jzaefferer/jquery-validation/blob/master/test/methods.js#L569 to support your new plugin

dkrahn:
I would suggest changing this part of the regular expression `(\(?[0-9]{2,3}\)?)?` for this `(\(0?[0-9]{2}\)|0?[0-9]{2})?` this way a strange number such as (09999999-9999 will not be allowed and the national long distance digit 0 is also validated.

staabm:
Closing because of missing tests. Feel free to add those, we can re-open again.

Issue #1567, opened 2015-09-01
Fully Researched, Bug-ish? Remote call just uses previous validation if own value has not changed


christopherbauer:
I'd like to preface this with the simple fact that this is SUCH an easy and straightforward issue that no jsfiddle needs to be provided, it's literally 3 lines of code that are probably the easiest 3 lines to read in the whole repo, so I'm just going to explain my issue and my rather plain way of resolving it...

As a developer
I want to have my zipcode validated for my state when I change the state
so that I can guarantee my data is valid

To do this, I wrote a custom remote rule:

            $("#zipcode").rules("add", { //when entering zipcode, check if is valid in state
                "remote": {
                    url: "zipcodeApiUrl",
                    type: "GET",
                    data: {
                        stateAbbreviation: function () {
                            return $("#state").val(); //send state code
                        }
                    }
                }
            });
            $("state").on("change", function () { //when changing state
                $("zipcode").valid(); //then check if zipcode is valid
            });

The offending line is 1320-1322 inside of the remote property of jquery.validate.js:

if ( previous.old === value ) {
     return previous.valid;
}

So I change my state from NY to OK but leave the zipcode (11735) in place, the first call to the api endpoint returns a bad validation result for OK and 11735 - perfect!

But then I realize I changed the state by mistake, so I go back and change it to NY...

But it tells me that the validation is still failing... for OK!

I suppose this is to reduce the number of calls to remote apis when the form is being validated. 

I would have sent up a patch for this, but that would really just have been me deleting lines 1320,1321, and 1322 :). However, this occurs to me as a breaking change that would (possibly) disrupt other people's workflow, in addition possibly adding an inefficiency.

If it were me, I'd just add an additional option for "perform check always" to the remote option (defaulted to false, of course), which I will do if someone gives me some positive feedback on this solution.

So if anyone would like to outline another option, or if the repo owner could give me a nudge in the right direction, I will go ahead and do it gladly (following the contrib guidelines), but I wanted to get some consensus before I go knocking down doors, especially since I can identify a clear use case in both directions.

christopherbauer:
I have a fix that shouldn't bother anyone AFAIK - https://github.com/christopherbauer/jquery-validation/tree/remotecheckpreviousdata. I'd love to pull request it but I'm at work and I'm blocked from using grunt and npm so I'll have to test later when I'm home :+1:

Arkni:
@christopherbauer 
This issues should be closed by now as your fix has been merged.

PR #1570, opened 2015-09-04
#1567 Instead of using validating field's value to skip api call, use the serialized data object of the request

christopherbauer:
Fix for #1567 ; Instead of just using validating field's value to determine whether or not to skip the remote check, use the data of the request that would go to the endpoint, as this may affect the validation result.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1291,7 +1291,7 @@ $.extend( $.validator, {
 			}
 
 			var previous = this.previousValue( element ),
-				validator, data;
+				validator, data, optionDataString;
 
 			if (!this.settings.messages[ element.name ] ) {
 				this.settings.messages[ element.name ] = {};
@@ -1300,12 +1300,12 @@ $.extend( $.validator, {
 			this.settings.messages[ element.name ].remote = previous.message;
 
 			param = typeof param === "string" && { url: param } || param;
-
-			if ( previous.old === value ) {
+			optionDataString = $.param( $.extend( { data: value }, param.data ) );
+			if (previous.old === optionDataString) {
 				return previous.valid;
 			}
 
-			previous.old = value;
+			previous.old = optionDataString;
 			validator = this;
 			this.startRequest( element );
 			data = {};
--- a/test/methods.js
+++ b/test/methods.js
@@ -470,6 +470,42 @@ asyncTest("remote extensions", function() {
 	strictEqual( v.element(e), true, "still invalid, because remote validation must block until it returns; dependency-mismatch considered as valid though" );
 });
 
+test("remote, data previous querystring", function() {
+	expect(4);
+	var change = true,
+		e = $("#username"),
+		v = $("#userForm").validate({
+			rules: {
+				username: {
+					required: true,
+					remote: {
+						url: "users.php",
+						type: "POST",
+						data: {
+							email: function() {
+								if ( change ) {
+									change = false;
+									return "email.com";
+								}
+								return "email2.com";
+							}
+						},
+						beforeSend: function(request, settings) {
+							if ( change ) {
+								deepEqual(settings.data, "username=asdf&email=email.com");
+							} else {
+								deepEqual(settings.data, "username=asdf&email=email2.com");
+							}
+						}
+					}
+				}
+			}
+		});
+	$("#username").val("asdf");
+	strictEqual( v.element(e), true, "new email value (email.com), new request; dependency-mismatch considered as valid though" );
+	strictEqual( v.element(e), true, "new email value (email2.com), new request; dependency-mismatch considered as valid though" );
+});
+
 module("additional methods");
 
 test("phone (us)", function() {


christopherbauer:
Hey guys, this is probably an obnoxious request, but can anyone help me with this phantomjs timeout issue? The change I made seemed pretty minor but now all the tests but one fail and phantomjs blows up. I'm not too familiar with npm/grunt and javascript testing in general, but I feel this change fixes a pretty major issue w/ remote validation.

christopherbauer:
Scratch that, I broke remote extensions test and got distracted by a weird error message.

christopherbauer:
Had my branches confused! Finally noticed the wrong line of code.

staabm:
Doesnt look that bad. Could you also add a unit test which covers the newly added functionality?

Do we have a test without data, to make sure we dont break existing functionality?

src/core.js
```diff
@@ -1300,12 +1300,12 @@ $.extend( $.validator, {
 			this.settings.messages[ element.name ].remote = previous.message;
 
 			param = typeof param === "string" && { url: param } || param;
-
-			if ( previous.old === value ) {
+			optionDataString = JSON.stringify( $.param( $.extend( { data: value }, param.data ) ) );
```
staabm: Is `$.param` really required? Wouldnt `stringify` already cover/support all we need?

christopherbauer: The $.param serializes the object into a query string, so that it can be stored as a property on the element itself (In the same way the Ajax request would send it to the api endpoint). This guarantees that jquery val will recognize when something about the request has changed instead of just the value of the validating field.

christopherbauer: I'd have to check if stringify would parse properly just on the object.

staabm: it converts a value into a json string, even objects and arrays:
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

christopherbauer: Yes and $.param converts to the url querystring, which is the 
representation that is sent to the api endpoint. That seems like the more 
correct way of determining whether or not a request was just recently sent 
to the endpoint, no?

staabm: oh I see, you dropped `stringify` for `$.param` which also works.

staabm:
patch looks good so far. open points are:

> Could you also add a unit test which covers the newly added functionality?
> 
> Do we have a test without data, to make sure we dont break existing functionality?

christopherbauer:
There already is an empty data test. I'll take a look at cobbling together a test ASAP, but I can't get NPM working at my job so it might take some time.

src/core.js
christopherbauer: Unless you think it's more correct to serialize the json object, I just pictured seeing a data property containing a JSON object and thought that would be bad or confusing vs seeing the querystring that was last sent to the endpoint, which makes more sense to me.

christopherbauer: Thanks for the help!

staabm: I am fine with either `stringify` or `$.param`, no preferences on my side.

only problem was in the initial version which used both in combination. Its fine as is.

christopherbauer:
Ok I added in some tests!

test/methods.js
```diff
@@ -470,6 +470,42 @@ asyncTest("remote extensions", function() {
...
+								if ( change ) {
+									change = false;
+									return "email.com";
+								}
+								return "email2.com";
+							}
+						},
+						beforeSend: function(request, settings) {
+							if ( change ) {
+								deepEqual(settings.data, "username=asdf&email=email.com");
```
staabm: this assertion tests a feature of jQuery, nothing related to jQuery-Validate? doesnt make sense in my eyes

staabm:
IMO we should have a test which asserts that a remote request is triggered, when the value of `data` changes.(Without your test this wasn't the case, because of the sortcut, which skipped the request.

christopherbauer:
I think we've got our wires crossed on what the purpose of this patch is. I'll try to break it down with a good ol' given-when-then:

given the value of one of the data parameters has changed
when validating
then should send a new remote validation api request

In the old code when the validating element's value changed, the remote methods sent a new request, else it would just return the previous validation result. That's bad because you could have a validation result be dependent on 30 data object properties and all of them could change, but if the element you called validate on hadn't changed in value, then you always got the same validation result.

My change is that if either a data property or the element's value changes the api request should run and return a *new* validation status.

When the validation response is triggered is not the focus of this pull request, the focus is making sure THAT validation to a remote endpoint occurs given a value of any data property has changed.

test/methods.js
christopherbauer: I can remove these lines if they provide no value, but I put them there because I saw the same asserts being performed in pre-existing remote tests in the methods.js file.

staabm:
thats what I wrote, isn't it.

we should have a test which verifies/asserts that the remote request is triggered when the data provided via `.data` property changes. This is what this PR actually does, no?

I think we are on the same page.

test/methods.js
staabm: could you point me to the test you are referring to? I quickly scanned the method.js and couldn't find where this is the case.

christopherbauer:
Well, I'm not writing a hook against the data parameters that if the value of the data parameters changes a validation request is sent; I wrote a patch that when you do something that causes a validation event to occur, the remote rule doesn't take a shortcut if the data parameters have changed at some other point. I left the actual activation of validation up to the end-user.

Example on how I foresaw this being used:

            $("#state").on("change", function() {       //when changing state...
                $("#zipcode").valid();                          //...re-validate zipcode (up to the developer WHEN this occurs
            });
            $("#zipcode").rules("add", {
                "remote": { //I revalidate even if ONLY the stateAbbreviation has changed...
                    url: "zipcodeApiUrl",
                    type: "GET",
                    data: {
                        stateAbbreviation: function() {
                            return $(stateId).val();
                        }
                    }
                }
            });

staabm:
thats exactly what I said. with your change we now get a remote-request via `$.ajax` when the contents of the `data` property changes - which was not the case before your patch.
the value of the element itself was already taken into account before.

This is the behaviour we need to test.

christopherbauer:
(sorry I can't seem to get the line comments working right here)

Is that not what these strictEqual calls are doing at the end of the test? The username textbox value becomes "asdf", then we call v.element(e) which validates the first request contains the data-property value 'email.com'.

+	strictEqual( v.element(e), true, "new email value (email.com), new request; dependency-mismatch considered as valid though" );

Then right after that, we validate again but this time we're looking for the data parameter value to have changed to 'email2.com' (because the data-property changes on the second request to 'email2.com':

+	strictEqual( v.element(e), true, "new email value (email2.com), new request; dependency-mismatch considered as valid though" );

christopherbauer:
Oh, you know what, this is an issue that I'm not familiar with qunit. I'm sorry, I see what you're asking. That's what I get for working on this so late.

staabm:
IMO the test should look similar to (untested):

```
test("remote, validation re-triggered when dependent fields change", function() {
	expect(4);
	var succeeded = 0,
		$f = $("#firstnamec"),
		v = $("#testForm1clean").validate({
			rules: {
				lastname: {
					remote: {
						url: "users.php",
						type: "POST",
						data: {
							firstname: function() {
								return $f.val();
							}
						},
						success: function(request, settings) {
							succeeded++;
						}
					}
				}
			}
		});
        strictEqual(succeeded, 0, "no initial request");
        v.valid();
        strictEqual(succeeded, 1, "remote request triggered on initial validation");
        v.valid();
        strictEqual(succeeded, 1, "no additonal request when nothing changed");
        $f.val('new-firstname');
        v.valid();
        strictEqual(succeeded, 2, "remote request triggered, because dependend field changed");
});
```

christopherbauer:
Ok thanks a lot! I'll take a look when I can get on a non-firewall'd connection.

christopherbauer:
An update; I spent about 3 hours on this and got nowhere. I have no idea why it's so difficult. I must be mistyping something somewhere. Does this look right to you?



test("remote, data previous querystring", function() {
	expect(4);
	var succeeded = 0,
		$f = $("#firstnamec"),
		$l = $("#lastnamec"),
		v = $("#testForm1clean").validate({
			rules: {
				lastnamec: {
					remote: {
						url: "users.php",
						type: "POST",
						data: {
							firstname: function() {
								return $f.val();
							}
						},
						success: function(request, settings) {
							succeeded++;
							alert(succeeded);
						}
					}
				}
			}
		});
	$f.val("first-name");
	$l.val("last-name");
	strictEqual( succeeded, 0, "no validation means no successful validation" );
	v.valid();
	strictEqual( succeeded, 1, "should validate first time" );
	v.valid();
	strictEqual( succeeded, 1, "should return same value as previous" );
	$f.val("different-first-name");
	v.valid();
	strictEqual( succeeded, 2, "should validate second time" );
});

christopherbauer:
Got it!

staabm:
@christopherbauer thanks, looks good to me.
I run the test and it fails without your patch which is great.

Would be awesome if you could squash the commits to 1, then we are ready to land this patch.
Thanks so far.

christopherbauer:
Ok, squashed it down to 1 commit and 1 message. I'm not sure if you're an official maintainer/owner of jquery-validation, but thanks for the guidance on this @staabm!

staabm:
@christopherbauer please prefix the commit with "Core:" not "Content". I will merge after this small tweak thx.

christopherbauer:
We're good to go!

staabm:
thanks @christopherbauer

christopherbauer:
Hey @staabm somehow the wrong version of methods.js ended up getting squashed and merged. I can't seem to recover the file, do you still have it?

PR #1572, opened 2015-09-08
Allow non-singleton usage.

geoidesic:
This is useful e.g. if you want to apply conditional validation to a form. Such as a form where parts are hidden (and thus should not be validated) under certain conditions . This is a backwards-compatible change which triggered by an option.

To make use of it just add the option: {singleton: false}. That way the object will be re-initialised on every validation call; thus allowing you to pass different options in with each call.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -10,11 +10,14 @@ $.extend($.fn, {
 			return;
 		}
 
-		// check if a validator for this form was already created
-		var validator = $.data( this[ 0 ], "validator" );
-		if ( validator ) {
-			return validator;
-		}
+		// Allow non-singleton:- useful e.g. if you want to apply conditional validation to a form
+		if (!options || typeof options.singleton === 'undefined' || options.singleton === true) {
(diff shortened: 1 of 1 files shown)


staabm:
please give more details on your use-case. hidden fields are not validated by default, so this cannot be the reason for your change.

in case you need to redefine all rules from scratch and dont want to rely on the current validators state you could also call the newly added `destroy` command[1].

I will close the issue, pleae feel free to discuss your points here nevertheless. I can re-open the issue again if required.

[1] https://github.com/jzaefferer/jquery-validation/pull/1392

geoidesic:
It wasn't about hidden fields (as in ```type="hidden"```) but rather about hidden HTML, as in ```display: none;```

However. I've just discovered a problem with my modification, which is that the ```submitHandler``` will keep adding events to the form, potentially creating a scenario where the form gets submitted multiple times. So perhaps your ```destroy``` method is a better idea.

Issue #1573, opened 2015-09-08
Bug in ver.1.14.0. input accepts string with whitespaces and validates to valid .


Konrud:
I'm using  **jquery.validate.js ver.1.14.0**.

If input/text-area has only *required* rule, it accepts string with whitespaces and validates to valid input.

I've tried this with  **jquery.validate.js ver.1.13.1** it works correct, string with whitespaces is not allowed.

Konrud:
I think the problem is in methods -> required method in its return statement,
now it's `return value.length > 0` but it accepts whitespaces, so `trim()` method should be added (`return $.trim( value ).length > 0;`) .

methods: required method, now it is looks like that :
```
	methods: {

		// http://jqueryvalidation.org/required-method/
		required: function( value, element, param ) {
			// check if dependency is met
			if ( !this.depend( param, element ) ) {
				return "dependency-mismatch";
			}
			if ( element.nodeName.toLowerCase() === "select" ) {
				// could be an array for select-multiple or a string, both are fine this way
				var val = $( element ).val();
				return val && val.length > 0;
			}
			if ( this.checkable( element ) ) {
				return this.getLength( value, element ) > 0;
			}
			return value.length > 0;
		},
```
And should be:
```
	methods: {

		// http://jqueryvalidation.org/required-method/
		required: function( value, element, param ) {
			// check if dependency is met
			if ( !this.depend( param, element ) ) {
				return "dependency-mismatch";
			}
			if ( element.nodeName.toLowerCase() === "select" ) {
				// could be an array for select-multiple or a string, both are fine this way
				var val = $( element ).val();
				return val && val.length > 0;
			}
			if ( this.checkable( element ) ) {
				return this.getLength( value, element ) > 0;
			}
			return $.trim( value ).length > 0;
		},
```

staabm:
This changed intentionally, see https://github.com/jzaefferer/jquery-validation/commit/2ad0ca82daf091fc692018b38e01e896fd304015

Konrud:
But now input accepts string with whitespaces.

PR #1574, opened 2015-09-08
Update core.js

Konrud:
Without trim() method, in methods: required method, input accepts string with whitespaces.


staabm:
Discussion at https://github.com/jzaefferer/jquery-validation/issues/1573#issuecomment-138669860

staabm:
As the initial change was intended, I dont see a way to land this.

PR #1578, opened 2015-09-11
Core: method.js fix tests for #1567

christopherbauer:
Incorrect tests were rebased.


christopherbauer:
@staabm Sorry for all the mess, I rebased from a system that I hadn't issued a pull request from and I temporarily lost the tests. I was able to find them and bring them back on - this should be the last fix.

staabm:
Thanks for the followup

PR #1579, opened 2015-09-14
Fixes #1508: Validation fails to trigger when next field is already filled out

### LINKED ISSUE #1508: Validation fails to trigger when next field is already filled out
To reproduce:

On the index demo, edit the markup to fill in the second field with a valid email address. Load that demo, type "abc" into the first field, tab to the next field, tab back, delete the value.

Expected: "This field is required" error.
Actual: No error

Originally reported as #244, but its really not specific to select elements. Reverted the fix from #396 in 5934f4d4509b900594b40eb4399446dd46debf36, as the caused a regression (#521).

This seems pretty tough to fix, as it scratches on the very basics of how the plugin figures out if a field should be validated or not. Haven't yet been able to come up with a better solution.

Arkni:
See #1508 for more information.

This pull request removes also the need for `lastElement`

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -272,7 +272,7 @@ $.extend( $.validator, {
 
 			if ( event.which === 9 && this.elementValue( element ) === "" || $.inArray( event.keyCode, excludedKeys ) !== -1 ) {
 				return;
-			} else if ( element.name in this.submitted || element === this.lastElement ) {
+			} else if ( element.name in this.submitted || this.isValidElement( element ) ) {
 				this.element( element );
 			}
 		},
@@ -410,8 +410,6 @@ $.extend( $.validator, {
 				checkElement = this.validationTargetFor( cleanElement ),
 				result = true;
 
-			this.lastElement = checkElement;
-
 			if ( checkElement === undefined ) {
 				delete this.invalid[ cleanElement.name ];
 			} else {
@@ -420,7 +418,7 @@ $.extend( $.validator, {
 
 				result = this.check( checkElement ) !== false;
 				if ( result ) {
-					delete this.invalid[ checkElement.name ];
+					this.invalid[ checkElement.name ] = false;
 				} else {
 					this.invalid[ checkElement.name ] = true;
 				}
@@ -466,7 +464,6 @@ $.extend( $.validator, {
 				$( this.currentForm ).resetForm();
 			}
 			this.submitted = {};
-			this.lastElement = null;
 			this.prepareForm();
 			this.hideErrors();
 			var i, elements = this.elements()
@@ -492,7 +489,9 @@ $.extend( $.validator, {
 			var count = 0,
 				i;
 			for ( i in obj ) {
-				count++;
+				if ( obj[ i ] ) {
+					count++;
+				}
 			}
 			return count;
 		},
@@ -508,6 +507,15 @@ $.extend( $.validator, {
 
 		valid: function() {
 			return this.size() === 0;
+		},
+
+		// Check if the given element is valid
+		// return
+		//          true  If the field is valid
+		//         false  If the field is invalid
+		//     undefined  If the field is not validated yet.
+		isValidElement: function( element ) {
+			return this.invalid[ element.name ] === undefined ? undefined : !this.invalid[ element.name ];
 		},
 
 		size: function() {
--- a/test/test.js
+++ b/test/test.js
@@ -1360,7 +1360,7 @@ test( "#1508: Validation fails to trigger when next field is already filled out"
 	element.trigger( "keyup" );
 
 	// element should be invalid
-	equal( false, validator.isValidElement( element[0] ), "The element is not valid" );
+	equal( validator.isValidElement( element[0] ), false, "The element is not valid" );
 });
 
 test( "validate checkbox on click", function() {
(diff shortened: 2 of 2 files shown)


Arkni:
@jzaefferer 

Can you please take a look at this (If you have the time of course).

Arkni:
@staabm 

Could you please review this PR ?

staabm:
@Arkni did you make sure that the unit test fails without your patch?

Arkni:
@staabm 
Yes, I did.

Here it is:
```js
test( "#1508: Validation fails to trigger when next field is already filled out", function() {
	// The next field is already filled out.
	$( "#lastname" ).val( "something" );

	var event     = $.Event( "keyup", { keyCode: 9 } ),
		element   = $( "#firstname" ),
		validator = $( "#testForm1" ).validate();

	// Fill the first element
	element.val( "abc" );

	// Tab to the next field
	element.blur();
	$("#lastname").trigger( event );
	$("#lastname").focus();

	// Tab back to element
	$("#lastname").blur();
	element.trigger( event );
	element.focus();

	// Delete the content
	element.val( "" );
	element.trigger( "keyup" );

	// element should be invalid
	equal( 1, validator.numberOfInvalids(), "The element is not valid" );
});
```

And it fails as expected without my patch.

```
Running "qunit:files" (qunit) task
Testing test/index.html ............................................................................F....................................................................................................................................
>> events - #1508: Validation fails to trigger when next field is already filled out
>> Message: The element is not valid
>> Actual: 1
>> Expected: 0
>> at file:///home/arkni/Git%20Repositories/jquery-validation/test/qunit/qunit.js:1311
>>   at file:///home/arkni/Git%20Repositories/jquery-validation/test/qunit/qunit.js:1907
>>   at file:///home/arkni/Git%20Repositories/jquery-validation/test/test.js:1363
>>   at file:///home/arkni/Git%20Repositories/jquery-validation/test/qunit/qunit.js:895
>>   at file:///home/arkni/Git%20Repositories/jquery-validation/test/qunit/qunit.js:1024
>>   at process (file:///home/arkni/Git%20Repositories/jquery-validation/test/qunit/qunit.js:583)
>>   at begin (file:///home/arkni/Git%20Repositories/jquery-validation/test/qunit/qunit.js:628)
>>   at file:///home/arkni/Git%20Repositories/jquery-validation/test/qunit/qunit.js:644

Warning: 1/1220 assertions failed (5338ms) Use --force to continue.
```

Is this the right way to test it ?

src/core.js
```diff
@@ -420,7 +418,7 @@ $.extend( $.validator, {
 
 				result = this.check( checkElement ) !== false;
 				if ( result ) {
-					delete this.invalid[ checkElement.name ];
+					this.invalid[ checkElement.name ] = false;
```
Arkni: Instead of deleting this property, I assign to it `false` to refer to valid element. So:
`false` => valid element
`true` => invalid element

src/core.js
```diff
@@ -510,6 +509,15 @@ $.extend( $.validator, {
 			return this.size() === 0;
 		},
 
+		// Check if the given element is valid
+		// return
+		//          true  If the field is valid
+		//         false  If the field is invalid
+		//     undefined  If the field is not validated yet.
+		isValidElement: function( element ) {
```
Arkni: I added also this function in order to check if the given element is valid or not. I don't know if it can be in use to others, but I used it in this patch.

src/core.js
Arkni: @jzaefferer 
What do you think ?

test/test.js
```diff
@@ -1334,6 +1334,35 @@ test( "don't revalidate the field when pressing special characters", function()
...
+	$("#lastname").blur();
+	element.trigger( event );
+	element.focus();
+
+	// Delete the content
+	element.val( "" );
+	element.trigger( "keyup" );
+
+	// element should be invalid
+	equal( false, validator.isValidElement( element[0] ), "The element is not valid" );
```
Arkni: **<< TO ME >>**
This should be changed to
```js
equal( validator.isValidElement( element[0] ), false, "The element is not valid" );
```

staabm:
I will double test your change on monday, thanks

Arkni:
@staabm 
Any chance to test this ?

staabm:
@Arkni just tested your changes locally.

I am running into the following error, in Chrome-Beta and FF-Dev-Edition:
{code}
configuration with attributes : #1508: Validation fails to trigger when next field is already filled out (1, 0, 1)Rerun13 ms
Died on test #1     at file:///C:/Users/mstaab/Documents/GitHub/jquery-validation/test/test.js:1856:1: validator.isValidElement is not a function@ 13 ms
Source: 	
TypeError: validator.isValidElement is not a function
    at Object.<anonymous> (file:///C:/Users/mstaab/Documents/GitHub/jquery-validation/test/test.js:1882:19)
{code}

staabm:
@Arkni just tested your changes.
- manually tested that issue #1508 fails without your test and works with your patch as expected
- unit test looks fine

thanks.

src/core.js
staabm: having another look at this method...

it assumes that `validate()` was called before, right? I think we should at least mention this precondition in the method doc.

Arkni: Yes, you're right.

`validate()` should be called before.

PR #1584, opened 2015-09-23
Core: The remote validation removes highlight from other invalid fields.

codezzly:
This Pull Request fixes problem with clearing the highlight in invalid fields after remote validation. When the remote validation receives response with false value, adds the highlight for field and at the same time removes the highlight from other invalid fields, that were previously validated. This situation occurs during autofilling fields by browser, in Chrome for example.

Example in fiddle:
http://jsfiddle.net/wc862mn2/
Step by step:
Please fill only the field "Phone number" and click "Send me". After the form validation, all labels will appear under fields and only the field with phone number will have highlight. Others required fields do not have highlight.

The same example with my fix:
http://jsfiddle.net/wc862mn2/1/


staabm:
@Arkni any opinions on this fix?

src/core.js
```diff
@@ -441,7 +441,6 @@ $.extend( $.validator, {
 			if ( errors ) {
 				// add items to error list and map
 				$.extend( this.errorMap, errors );
-				this.errorList = [];
```
Arkni: This can be a solution, but I'm not sure right now.

Arkni:
@codezzly 
Could you please add a unit test ?

src/core.js
staabm: wouldnt this lead to `errorList` growing and growing, e.g. when calling `showErrors` twiceß

Arkni: Yes, you're right. That's why I said `I'm not sure`.

I think we should only extend `this.errorList`, so we have to override old errors and add new ones. Right now I'm thinking of doing:
```js
this.errorList = $.grep( this.errorList, function( error ) {
	return !( error.element.name in errors );
} );
```
in place of set an empty array to `this.errorList`

What do you think ?

staabm: still not sure this is the right thing todo. could we clear the error list and add all elemens from the `errorMap` into the list? I mean the errorMap contains everything but the errorList doesn't, right?

Arkni: > could we clear the error list and add all elemens from the errorMap into the list? I mean the errorMap contains everything but the errorList doesn't, right?

That seems the right way. Yes the errorMap contains all needed errors, we only have to map it to an array and assign the result to `this.errorList`

> I am wondering that we can see all validation error messages on the fields but only the invalid indicators/borders are missing...

Yes, only the highlight was removed. This is caused due to `valideElements` return all elements except the one the remote rule was attached to. ( see [core.js#L796-L800](https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L796-L800) and [core.js#L806-L814](https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L806-L814) )

staabm: I also dont get why we have a `errorMap` and a `errorList`. Those are event handed over to library consumers in `validator.settings.showErrors` callbacks, see http://jqueryvalidation.org/validate/#showerrors

Arkni: Maybe it's time to remove one of them.

staabm: > Maybe it's time to remove one of them.

cannot be done, without a bc break I guess.

Arkni: > cannot be done, without a bc break I guess.

Yes, it's a BC break. Let's discuss that in a new issue

About this pull request, and as per your suggestion above, we can do the following:
```diff
diff --git a/src/core.js b/src/core.js
index c85669a..a78d457 100644
--- a/src/core.js
+++ b/src/core.js
@@ -451,16 +451,16 @@ $.extend( $.validator, {
 		// http://jqueryvalidation.org/Validator.showErrors/
 		showErrors: function( errors ) {
 			if ( errors ) {
+				var validator = this;
 
 				// Add items to error list and map
 				$.extend( this.errorMap, errors );
-				this.errorList = [];
-				for ( var name in errors ) {
-					this.errorList.push( {
-						message: errors[ name ],
-						element: this.findByName( name )[ 0 ]
-					} );
-				}
+				this.errorList = $.map( this.errorMap, function( message, name ) {
+					return {
+						message: message,
+						element: validator.findByName( name ) [ 0 ]
+					};
+				} );
 
 				// Remove items from success list
 				this.successList = $.grep( this.successList, function( element ) {
```

staabm: Did test this change against the fiddle provided by @codezzly if it fixes our problem?

Arkni: > Did test this change against the fiddle provided by @codezzly if it fixes our problem?

Yes, I tested it against the provided fiddle and locally as well. :white_check_mark:

staabm: @Arkni awesome thx. so only thing left is a unit test, lets see If I get something crafted together

staabm: > Yes, it's a BC break. Let's discuss that in a new issue

agree

Arkni: > so only thing left is a unit test, lets see If I get something crafted together

Yes, we can do that. But lets wait for a feedback from @codezzly

codezzly: Hello, thanks for your interest in my issue. I tested your code and I can tell that it works perfectly! Thank you so much for help. Could you merge this fix in near future?

codezzly:
@Arkni yes, of course. I can write a unit test for this issue but I need help from you. Could you tell me what should check in this case test? To which file I should add this test? Please, tell me little guidance.

Arkni:
>Could you tell me what should check in this case test?

Normally, you should translate the steps -- you provided in first comment -- into code,so you have to test that the other elements (firstname and lastname) have the `error` class, which should fail without your fix and passed otherwise. But before that, you have to update your fix with the one I provided in [#1584 (comment)](https://github.com/jzaefferer/jquery-validation/pull/1584#discussion_r44677327).

>To which file I should add this test?

You add the unit test to [/test/methods.js](https://github.com/jzaefferer/jquery-validation/blob/master/test/methods.js), you can find some unit tests of the remote method that you can base on them to create a unit test for this PR.

If you need any help, just ping me :)

staabm:
@Arkni we agreed that the fix for this PR should be like the code in https://github.com/jzaefferer/jquery-validation/pull/1584#discussion_r44677327,right?

Arkni:
> @Arkni we agreed that the fix for this PR should be like the code in #1584 (comment)?

I guess yes, and @codezzly already tested it:
>>Hello, thanks for your interest in my issue. I tested your code and I can tell that it works perfectly!

Also, we was thinking about removing the `errorList` as `errorMap` contains all needed errors.

Arkni:
@codezzly any update on this?

codezzly:
Hi, I updated my PR. Thanks to @Arkni for your help and provide of good fix. I added test of this PR in methods.js - please for you to check this, I hope that my first test is correct.

src/core.js
```diff
@@ -459,16 +459,16 @@ $.extend( $.validator, {
...
-				for ( var name in errors ) {
-					this.errorList.push( {
-						message: errors[ name ],
-						element: this.findByName( name )[ 0 ]
-					} );
-				}
+				this.errorList = $.map( this.errorMap, function( message, name ) {
+					return {
+						message: message,
+						element: validator.findByName( name ) [ 0 ]
```
Arkni: There is an extra space between `)` and `[` in this line.

Arkni:
Other than my comment above, the PR LGTM :)

Let's see what @staabm think abut it.

src/core.js
Arkni: Just checked, it was my fault :(

staabm:
let me take care of this.

staabm:
@codezzly thanks.

PR #1586, opened 2015-09-24
Adding mechanism for custom asynchronous validation methods

gregorej:
Hi!

These are my changes that should enable any jquery-validation plugin users to create their own 'remote' validation method. This might be needed if jquery-validation plugin is to be used with backend that requires some additional authentication (tokens, passwords, etc.). generally, some non-standard AJAX communication. Example usage:

```javascript

myRemote: function( value, element, param ) {
			var data = {};
			data[ element.name ] = value;
			return this.callbackMethod(value, element, function(processResult) {
				$.ajax( {
					url: param,
					headers: {
						"Authentication": "MyToken"
					},
					dataType: "json",
					type: "post",
					data: data,
					success: function( response ) {
						processResult( response.status === "VALID", response.message );
					}
				} );
			}, data);
		}
     
```

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -899,6 +899,58 @@ $.extend( $.validator, {
 			return !$.validator.methods.required.call( this, val, element ) && "dependency-mismatch";
 		},
 
+		//can be used to create custom asynchronous validation methods (such as remote)
+		//The method parameter is a function that accepts single argument: callback function that should be used to
+		//handle result that arrived from asynchronous operation. Discriminator is any value that distinguishes two
+		//calls for same element (for example with different values).
+		//See remote method as an example usage.
+		callbackMethod: function(value, element, method, discriminator) {
+			function processResult(valid, message) {
+				var errors,  submitted;
+
+				validator.settings.messages[ element.name ].remote = previous.originalMessage;
+				if ( valid ) {
+					submitted = validator.formSubmitted;
+					validator.prepareElement( element );
+					validator.formSubmitted = submitted;
+					validator.successList.push( element );
+					delete validator.invalid[ element.name ];
+					validator.showErrors();
+				} else {
+					errors = {};
+					if (message[ "default" ]) {
+						message = validator.defaultMessage( element, message[ "default" ] );
+					}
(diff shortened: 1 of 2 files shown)


staabm:
The builtin-in `remote` validator already allows you to pass in arbitrary options, to be passed into the underlying `$.ajax` call, see https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L1341

In case you need auth tokens or similar to contact your backend, you maybe should even use `jQuery.ajaxSetup` to do define it for all requests hitting your backend, instead of passing those parameters arround everywhere.

gregorej:
OK, probably I need to be more precise about our case. ``ajaxSetup`` is not a solution for us becase before each request we need to first do an AJAX request for a authentication token. So it goes like this:

```javascript
   getToken({
       success: function(token) {
            $.ajax({
                 url: '/validation/endpoint',
                 headers: {
                      'Authentication': token
                 }
                 success: function(response) {
                          //do something with response
                 }
         }
})
```

So in fact our ajax call is two calls: first for token, second for the real data (wtih token). Current ``remote`` as far I understand does not support this.

staabm:
@gregorej so you are saying that you always need 2 ajax calls, everytime?

gregorej:
Pretty much yes, the token is valid for some time but to make sure we are up to date we do 2 calls. I am not saying it is perfect, it is just the way the backend works.

I was thinking the solution I propose would enable implementing some more complex async validations like ours.

staabm:
since your use-case is rather specific I dont think we need this functionality in the jquery-validation core.

Furthermore I think client-side API should not need to be aware of your authenticiation mechanism, otherwise you need to tackle this very same issue for every client side lib which does ajax requests to your backend.

Couldn't you set a timer/timeout on the client which automatically re-news your access tokens once invalid and use `jQuery.ajaxSetup` to pass in this token for all your backend requests. This would also have the benefit that you dont have to do the authentication calls in that polling-like manner.

gregorej:
Polling for token sounds like an option. I will consider it. Thanks!

PR #1587, opened 2015-09-24
Fix faulty aria-describedby value

Nettsentrisk:
The error ID was being escaped for use in a regex, but these escaped characters were also being passed on to the aria-describedby value set in the HTML, which means the ID doesn't match up with the ID of the element it's referring to.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -795,16 +795,16 @@ $.extend( $.validator, {
 				} else if ( error.parents( "label[for='" + elementID + "']" ).length === 0 ) {
 					// If the element is not a child of an associated label, then it's necessary
 					// to explicitly apply aria-describedby
-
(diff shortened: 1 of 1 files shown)


Arkni:
I think there is something wrong with your pull request.
_0 files changed_ :open_mouth:

staabm:
Fixed by https://github.com/jzaefferer/jquery-validation/commit/b2df81ccb6276a6a39661ebd1ddd4a54c182217e thanks.

staabm:
hmm actually I am not sure, this is correct.

@Nettsentrisk could you provide a jsfiddle or at best a unit test which reproduces the error you are referring to?

Nettsentrisk:
Don't have time to make one now, but it will fail if there's an input field with a name that contains a period, for example "Test.Name". The ID for the element referred to by aria-describedby will then become "Test\.Name" and thereby not be the same as the ID in aria-describedby.

PR #1588, opened 2015-09-25
Revert "Core: escape errorID only for use in the regex"

staabm:
This reverts commit b2df81ccb6276a6a39661ebd1ddd4a54c182217e.


staabm:
The suggested change breaks the built. lets wait until we have a unit test for the error reported in #1587

PR #1592, opened 2015-09-25
Gruntfile.js: fix for require exports

Badalik:


staabm:
@Badalik whats the background of the change? Why is it required, what is the issue?

I dont have experience with grunt so, I cant merge it without enough information from your side.

@Arkni maybe you can have a 2nd look at it.

.gitignore
```diff
@@ -1,2 +1,4 @@
-dist
+#dist
```
Arkni: Please revert this change.

Arkni:
You should not include the dist folder (and its files), please remove them and then push again.

Gruntfile.js
```diff
@@ -25,7 +25,9 @@ banner = "/*!\n" +
 umdStart = "(function( factory ) {\n" +
 	"\tif ( typeof define === \"function\" && define.amd ) {\n";
 
-umdMiddle = "\t} else {\n" +
+umdMiddle = "\t} else if (typeof exports === \"object\") {\n" +
```
Arkni: I think you want to add CommonJS pattern. If that so, you have to change this line to:
```js
} else if ( typeof module === 'object' && module.exports ) {
```

Gruntfile.js
```diff
@@ -25,7 +25,9 @@ banner = "/*!\n" +
...
+	"\t\tfactory(require(\"jquery\"));\n" +
```
Arkni: This line should be changed to:
```js
module.exports = factory( require( "jquery" ) );
```

Badalik:
I work with a template on which EC6 then use babel.js transforms in EU5. We use import EC6 and npm require and without the code plugins are not working.

Badalik:
You can make it in his own way, I just have a problem. This is a useful editing

Arkni:
@Badalik 
Did you try to test the suggested changes ?

staabm:
as far as I can see we already have landed a similar path with PR #1621.

please shout out if I miss something.

PR #1596, opened 2015-09-29
stopRequest() doesn't pass on the submit button value

das-peter:
The `stopRequest()` method currently doesn't properly handle cases in which the form was submitted (`validator.formSubmitted`) but no special submit handler (`validator.settings.submitHandler`) is defined.
The method triggers a submit if the form is valid, but as this is a scripted submit the value of the initially used submit button is missing.
If a submit handler is used the submit button is tracked and then injected as hidden field in `function handle() {`.

I propose we always track the submit button - no matter if `validator.settings.submitHandler`  is defined or not - and if `validator.settings.submitHandler` isn't used `stopRequest()` does its own handling of the submit button value.
Having `validator.submitButton` always set might be an API change - however as of now the library itself only uses the value if `validator.settings.submitHandler` is set too. Which means there should be no change in behavior despite always setting `validator.submitButton`.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -25,9 +25,9 @@ $.extend( $.fn, {
 		if ( validator.settings.onsubmit ) {
 
 			this.on( "click.validate", ":submit", function( event ) {
-				if ( validator.settings.submitHandler ) {
-					validator.submitButton = event.target;
-				}
+				// Track the used submit button to properly handle scripted
+				// submits later.
+				validator.submitButton = event.target;
 
 				// allow suppressing validation by adding a cancel class to the submit button
 				if ( $( this ).hasClass( "cancel" ) ) {
@@ -925,7 +925,20 @@ $.extend( $.validator, {
 			}
 			delete this.pending[ element.name ];
 			if ( valid && this.pendingRequest === 0 && this.formSubmitted && this.form() ) {
+				var hidden;
+				// If no submit handler is set but a submit button is present
+				// ensure the following scripted submit has the value of the
+				// submit button by injecting a hidden field.
+				if ( !this.settings.submitHandler && this.submitButton ) {
+					hidden = $("<input type='hidden'/>")
+					  .attr("name", this.submitButton.name)
+					  .val($(this.submitButton).val())
+					  .appendTo(this.currentForm);
+				}
 				$( this.currentForm ).submit();
+				if ( hidden ) {
+				  hidden.remove();
+				}
 				this.formSubmitted = false;
 			} else if ( !valid && this.pendingRequest === 0 && this.formSubmitted ) {
 				$( this.currentForm ).triggerHandler( "invalid-form", [ this ] );


staabm:
could you please post a jsfiddle which reproduces the reported issue (without your patch applied)

das-peter:
@staabm Thanks for the quick response - I'll see what I can do.
For now cross-linking to where this originally comes from: https://www.drupal.org/node/2576705

staabm:
@das-peter any news?

staabm:
@Arkni any idea on how to fix that issue?

Arkni:
@staabm I will investigate the issue when time allows, maybe this night :)

das-peter:
Finally found the time to create a fiddle:
https://jsfiddle.net/sofq4qbw/8/

If you hit one of the Submit Buttons you'll see that the custom submit handler is executed twice - which could be considered an issue on its own.
But more relevant is that the expected button value isn't available in the POST.
It displays this form data:
```
  "form": {
    "myfield": "test", 
  }, 
```
Compare it with this fiddle without the remote but a "normal" validator:
https://jsfiddle.net/zuqxLfey/2/
It displays this form data:
```
  "form": {
    "myfield": "test", 
    "operation": "Submit Two"
  }, 
```

Arkni:
@das-peter 

> If you hit one of the Submit Buttons you'll see that the custom submit handler is executed twice - which could be considered an issue on its own.

I think the second submit is legitimate, let's say that you have 3 fields,  one has the `remote` rule and the remaining 2 fields have the `required` rule.
If you hit the submit button, the plugin will validate all the elements, starting with the one with `remote` rule, then the remaining fields, at this rate there is one pending request. So the first submit event will be aborted due to this check inside `this.on( "submit.validate", function() { } )` inside the plugin:
>>**_`validator.form()` will return `true` even if there is a pending request, `remote` rule return `"pending"` for the first call, which is considered valid until the request finished._**

```js
if ( validator.form() ) {
	if ( validator.pendingRequest ) {
		validator.formSubmitted = true;
		return false;
	}
	return handle();
} else {
	validator.focusInvalid();
	return false;
}
```

After the request finished, the plugin call `stopRequest` method which will check the form again. If it is valid, the form will be submitted otherwise show form errors.

Now, let's say that there is no call to submit inside `stopRequest` method, in this case you will need to click the submit button a second time in order to submit your form, which is unacceptable to the end user.

----


> I propose we always track the submit button - no matter if validator.settings.submitHandler is defined or not - and if validator.settings.submitHandler isn't used stopRequest() does its own handling of the submit button value.
> Having validator.submitButton always set might be an API change - however as of now the library itself only uses the value if validator.settings.submitHandler is set too. Which means there should be no change in behavior despite always setting validator.submitButton.

+1 to this, but I'm not ok with the implementation. There is no need to duplicate the code to create the hidden element and append it. I think a proper implementation will be changing the `handle` function.
What do you think ?

Arkni:
Closed in favor of #2019 

The new PR is based on this one and the attribution is maintained.
Thanks for your contribution, though.

PR #1597, opened 2015-09-30
Build: support for commonjs exports

gregorej:
Inspired by [backbone's](https://github.com/jashkenas/backbone/blob/master/backbone.js#L24-L27) commonjs support


Arkni:
I suggest to do it like this: https://github.com/umdjs/umd/blob/master/jqueryPluginCommonjs.js#L10-L13

codeclown:
Having this merged would be great!

staabm:
@codeclown please adjust it per feedback above

codeclown:
Made a new pull request since not sure how to edit this one or if it's even desired.

https://github.com/jzaefferer/jquery-validation/pull/1621

Issue #1602, opened 2015-10-04
Adding transformation/normalization layer


Arkni:
The user can change/transform the value of an element before validating the element in question. The transformed value will be then used by the associated methods instead of the real one.

I'm thinking of adding something like this:
```js
$(".selector").validate({
    rules: {
        an_element: {
            required: true,
            othervalidator: true,
            // Transformer or another or better name
            transformer: function(element) {
                // Change/transforme the value of the element
                // And then return the transformed value
                return transformed_value;
            }
        }
    }
});
```

Arkni:
@staabm 
Can you delete our comments in issue-1601 , in order to let the thread clean ? and continue the discussion here

Arkni:
@jzaefferer 
Can you please take a look at this, and give us your feedback ?
Do you agree with adding this layer to the plugin ?

staabm:
I guess we should name this feature "normalizer" instead of transformer.

Or fan you think of wnuse case in which you actually transforms input into completely different values instead of just trimming spaces or stripping some special chars?

Arkni:
> I guess we should name this feature "normalizer" instead of transformer.

Sounds good to me, as `normalizer` reflects more what we are about to do.

> Or fan you think of wnuse case in which you actually transforms input into completely different values instead of just trimming spaces or stripping some special chars?

A better example is the one I mentioned in [#1615 (comment)](https://github.com/jzaefferer/jquery-validation/issues/1615#issuecomment-149241665), As the `C:\fakepath\` is not added in Firefox and _may_ change in Chrome and Edge in the coming releases. So instead of patching the plugin, the user can use the code from the spec instead.

Another example, is when using the `url` method, this require the user to enter at least `//` as in `//google.com` or to input the full url `tld://google.com` where `tld = https?|ftp`. So we can check if the used used the TLD and the `//` is not add them so the element can pass the check:
```js
$( "form" ).validate( {
    rules: {
        file_input: {
            required: true,
            url: true,
            normalizer: function( element ) {
                var url = element.value;

                // Check if it doesn't start with http:// or https:// or ftp://
                if ( url && url.substr( 0, 7 ) !== "http://"
                         && url.substr( 0, 8 ) !== "https://"
                         && url.substr( 0, 6 ) !== "ftp://" ) {
                    // then prefix with http:// or even https://
                     url = "http://" + url;
                }

                // Return the new url
                return url;
            }
        }
    }
} );
```

staabm:
[...]

PR #1609, opened 2015-10-09
Core: Added transformation/normalization layer

Arkni:
The user can change/transform the value of an element before validating
the element in question. The `normalizer value` will be then used by
the associated methods instead of the `real one`.

Ref: #1602

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -627,6 +627,22 @@ $.extend( $.validator, {
 				val = this.elementValue( element ),
 				result, method, rule;
 
+			// If a normalizer is defined for this element, then
+			// call it to retreive the changed value instead
+			// of using the real one.
+			// Note that `this` in the normalizer is `element`.
+			if ( typeof rules.normalizer === "function" ) {
+				val = rules.normalizer.call( element, val );
+
+				if ( typeof val !== "string" ) {
+					throw new TypeError( "The normalizer should return a string value." );
+				}
+
+				// Delete the normalizer from rules to avoid treating
+				// it as a pre-defined method.
+				delete rules.normalizer;
+			}
+
 			for ( method in rules ) {
 				rule = { method: method, parameters: rules[ method ] };
 				try {
@@ -861,7 +877,7 @@ $.extend( $.validator, {
 		// meta-characters that should be escaped in order to be used with JQuery
 		// as a literal part of a name/id or any selector.
 		escapeCssMeta: function( string ) {
-			return string.replace( /(!|"|#|\$|%|&|'|\(|\)|\*|\+|,|\.|\/|:|;|<|=|>|\?|@|\[|\\|\]|\^|`|\{|\||\}|~)/g, "\\$1");
+			return string.replace( /([\\!"#$%&'()*+,./:;<=>?@\[\]^`{|}~])/g, "\\$1" );
 		},
 
 		idOrName: function( element ) {
@@ -1104,7 +1120,7 @@ $.extend( $.validator, {
 
 		// evaluate parameters
 		$.each( rules, function( rule, parameter ) {
-			rules[ rule ] = $.isFunction( parameter ) ? parameter( element ) : parameter;
+			rules[ rule ] = $.isFunction( parameter ) && rule !== "normalizer" ? parameter( element ) : parameter;
 		} );
 
 		// clean number parameters
--- a/test/rules.js
+++ b/test/rules.js
@@ -305,3 +305,91 @@ test( "rules(), rangelength attribute as array", function() {
 		rangelength: [ 2, 3 ]
 	} );
 } );
+
+test( "rules(), normalizer", function() {
+	var username = $( "#usernamec" ),
+		urlc = $( "#urlc" ),
+		lastname = $( "#lastnamec" ),
+		v;
+
+	username.val( "\t\t \r" );
+	urlc.val( "" );
+
+	v = $( "#testForm1clean" ).validate( {
+		rules: {
+			username: {
+				required: true,
+				// Using the normalizer to trim the value of the element
+				// before validating it.
+				normalizer: function( value ) {
+					equal( this, username[ 0 ], "`this` in the normalizer should be the username element." );
+
+					// Trim the value of the input
+					return $.trim( value );
+				}
+			},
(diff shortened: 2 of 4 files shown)


staabm:
Will be on vacation for a week. Will review the PR when I am back.

Arkni:
Have a good holiday :)

staabm:
Please adjust the PR per discussion in https://github.com/jzaefferer/jquery-validation/issues/1602

Arkni:
Will do soon.

staabm:
No need to hurry :-)

Arkni:
@staabm 
Done!

Please let me know if there is anything I missed from our discussion.

test/rules.js
```diff
@@ -305,3 +305,58 @@ test( "rules(), rangelength attribute as array", function() {
...
+
+	var v = $( "#testForm1clean" ).validate( {
+		rules: {
+			username: {
+				required: true,
+				// Using the normalizer to trim the value of the element
+				// before validating it.
+				normalizer: function( element ) {
+					// Trim the value of the input
+					return $.trim( element.value );
```
staabm: Should this use `elementValue` instead?

test/rules.js
```diff
@@ -305,3 +305,58 @@ test( "rules(), rangelength attribute as array", function() {
...
+	$( "#usernamec" ).val( "\t\t \r" );
+	$( "#urlc" ).val( "" );
+
+	var v = $( "#testForm1clean" ).validate( {
+		rules: {
+			username: {
+				required: true,
+				// Using the normalizer to trim the value of the element
+				// before validating it.
+				normalizer: function( element ) {
```
staabm: Wdyt of passing in the elementValue instead of the element? We could change `this` to be the element instead...

src/core.js
```diff
@@ -624,8 +624,16 @@ $.extend( $.validator, {
...
 				} ).length,
 				dependencyMismatch = false,
-				val = this.elementValue( element ),
-				result, method, rule;
+				val, result, method, rule;
+
+			// If a normalizer is defined for this element, then
+			// call it to retreive the changed value instead
+			// of using the real one.
+			val = typeof rules.normalizer === "function" ? rules.normalizer.call( this, element ) : this.elementValue( element );
```
staabm: Not using a ternary could ease reading

staabm: Do we need to make sure that the normalizer returns a string?

test/rules.js
Arkni: I thought of passing two parameters, the `value` and the `element`, but if we set `this` to `element` it's fine to me.

test/rules.js
Arkni: See my answer above

src/core.js
Arkni: > Not using a ternary could ease reading

Should I use an `if` statement instead?

> Do we need to make sure that the normalizer returns a string?

The value can be of any type, I guess.

staabm: Yes, please use a if.

Technically it can return anything, but we require a string here?

Arkni: > Technically it can return anything, but we require a string here?

So, I need to check for the type of the returned value, and change `val` only if the type is `string`, right ?

staabm: I would throw a error if a normalizer returns something differen from a string, because it is a programmer mistake

Arkni: Make sense. Working on it.

Arkni:
@staabm 
Updated the PR.

staabm:
Nice. Only things missing
- a docs PR for the new feature at https://github.com/jzaefferer/validation-content
- a assertion in the test which makes sure `this` inside the normalizer is the correct DomElement as expected

Arkni:
> a assertion in the test which makes sure this inside the normalizer is the correct DomElement as expected

Done!

> a docs PR for the new feature at https://github.com/jzaefferer/validation-content

Will do that, this night I guess (or even tomorrow)

staabm:
@Arkni just realised we also need a test which makes sure a exception is thrown when normalizer returns a non string.

Arkni:
> @Arkni just realised we also need a test which makes sure a exception is thrown when normalizer returns a non string

I will see what I can do.

Arkni:
> @Arkni just realised we also need a test which makes sure a exception is thrown when normalizer returns a non string
> >I will see what I can do.

Done!

Arkni:
For the docs, I will create a PR when I have the time, maybe this weekend.

staabm:
Awesome, thx.

src/core.js
```diff
@@ -861,7 +877,7 @@ $.extend( $.validator, {
 		// meta-characters that should be escaped in order to be used with JQuery
 		// as a literal part of a name/id or any selector.
 		escapeCssMeta: function( string ) {
-			return string.replace( /(!|"|#|\$|%|&|'|\(|\)|\*|\+|,|\.|\/|:|;|<|=|>|\?|@|\[|\\|\]|\^|`|\{|\||\}|~)/g, "\\$1");
+			return string.replace( /([!"#$%&'()*+,./:;<=>?@\[\\\]^`{|}~])/g, "\\$1" );
```
Arkni: I have simplified the regex used in `escapeCssMeta`.

Arkni: Is it OK to include it in this PR ?

staabm: There migght be one `\` too much, othwise lgtm

Arkni: > There might be one \ too much

nope, one for `\` and one for `]`

staabm: Ok, so we should order the escaped `\` char somewhere where it doesnt follow directly another escape char. This would ease reading

Arkni: make sense. maybe like this:
```
/([\\!"#$%&'()*+,./:;<=>?@\[\]^`{|}~])/g
```

staabm: Looks good.

Arkni: Done!

Arkni:
> a docs PR for the new feature at https://github.com/jzaefferer/validation-content

See jzaefferer/validation-content#15

staabm:
Awesome sauce :star2:

Arkni:
:smiley:

Issue #1506, opened 2015-06-25
Should a good idea split this project?


vmariano:
Hi. I used this plugin in a couple of projects and help me to save lot of time. Thanks.

Very often in my the projects (with exception of ´required´) the build-in rules are not required, and i must add a lot of custom rules.

So, I think that should be a good idea split this project in 2. one with the implementation of the form validation core. And other repo/lib/etc (whatever) with plugins required to the project.

If is this is planed for the  next versions, ignore this issue.

Thanks.

tsjoberg:
+1 - especially if you only use 1 or 2 methods and dont need the other built-in methods

jzaefferer:
This isn't planned. I'm happy to review and merge pull requests to move methods from `core.js` to `additionals`, for example `creditcard` really shouldn't be in core.

PR #1612, opened 2015-10-12
Core: Split out `creditcard` method

Arkni:
As mentioned by Jörn in [#1506 (comment)](https://github.com/jzaefferer/jquery-validation/issues/1506#issuecomment-116327217)

### DIFF
--- a/src/additional/creditcard.js
+++ b/src/additional/creditcard.js
@@ -1,0 +1,40 @@
+// http://jqueryvalidation.org/creditcard-method/
+// based on http://en.wikipedia.org/wiki/Luhn_algorithm
+$.validator.addMethod( "creditcard", function( value, element ) {
+	if ( this.optional( element ) ) {
+		return "dependency-mismatch";
+	}
+
+	// accept only spaces, digits and dashes
+	if ( /[^0-9 \-]+/.test( value ) ) {
+		return false;
+	}
+
+	var nCheck = 0,
+		nDigit = 0,
+		bEven = false,
+		n, cDigit;
+
+	value = value.replace( /\D/g, "" );
+
+	// Basing min and max length on
+	// http://developer.ean.com/general_info/Valid_Credit_Card_Types
+	if ( value.length < 13 || value.length > 19 ) {
+		return false;
+	}
+
+	for ( n = value.length - 1; n >= 0; n-- ) {
+		cDigit = value.charAt( n );
+		nDigit = parseInt( cDigit, 10 );
+		if ( bEven ) {
+			if ( ( nDigit *= 2 ) > 9 ) {
+				nDigit -= 9;
+			}
+		}
+
+		nCheck += nDigit;
+		bEven = !bEven;
+	}
+
+	return ( nCheck % 10 ) === 0;
+}, "Please enter a valid credit card number." );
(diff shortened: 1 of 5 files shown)


staabm:
In case we have unit tests for this method, please move them into `test/additional/creditcard.js`.

In case we dont have any, would be great you could add simple coverage

Arkni:
> In case we have unit tests for this method, please move them into test/additional/creditcard.js.

Yes, there is a unit test for this method and it's under `methods` modules in `test/methods.js`
There is also a module named `additional methods` [(test/methods.js#L520)](https://github.com/jzaefferer/jquery-validation/blob/master/test/methods.js#L520) in the same file that contains the test for other additional methods.

I think I have to move all test for additional method to be under `additional methods` module. Or do I have to create a new file named `additional-methods.js` that will contains all test for the availabale additional methods ?

staabm:
I would expect to have one test file for each source file. Each testfile should contain only tests which cover the code in the corresponding src file. I guess this should be the long/mid-term goal. 

Please dont touch tests other then those related to creditcard within this PR though.

Arkni:
> I would expect to have one test file for each source file. Each testfile should contain only tests which cover the code in the corresponding src file. I guess this should be the long/mid-term goal.

Sounds good to me, but this will required assembling all testfiles into one named `additional-methods.js` or just append them to `methods.js`. So it will be easy to include them in `test.html`

Your thoughts ?

staabm:
@Arkni I am fine with anlong list of includs, dont need a file just for the purpose to add a unnecessary indirection

staabm:
I mean we only need to list them here, right?
https://github.com/jzaefferer/jquery-validation/blob/master/test/index.html#L14

Arkni:
> @Arkni I am fine with a long list of includes.. dont need a file just for the purpose to add a unnecessary indirection, because performance is not a primary concern in tests

make sense.

> I mean we only need to list them here, right?
> https://github.com/jzaefferer/jquery-validation/blob/master/test/index.html#L14

I will do that for `creditcard` method and in another PR for other additional methods

staabm:
> and in another PR for other additional methods
Yes we can do this in the same stroke when moving code of additionals away from core.js

Arkni:
There is a problem.

All method tests are depend on a method named [`methodTest`](https://github.com/jzaefferer/jquery-validation/blob/master/test/methods.js#L3-L12), which is defined inside a closure in `methods.js`.

I don't want to replicate the code.
Your thoughts please ?

staabm:
We only need to include our new test files after the method.js in the index.html

Arkni:
> I guess we need to move the methodTest into the global scope for now, so outside the closure.

Or event better by removing [methods.js#L1](https://github.com/jzaefferer/jquery-validation/blob/master/test/methods.js#L1)  and the last one [methods.js#L1276](https://github.com/jzaefferer/jquery-validation/blob/master/test/methods.js#L1276).

staabm:
Not sure why we have this closure their in the first place. I am ok with dropping it in case test suite stays green

Arkni:
> Not sure why we have this closure their in the first place. I am ok with dropping it in case test suite stays green

Done and test suite is green :white_check_mark:

staabm:
Looks perfect, thx. Will land this PR after we got some docs for it.

staabm:
Tanks man. :fireworks:

Arkni:
My pleasure :)

nebrot:
The creditcard-method is now in the wrong list [docs](http://jqueryvalidation.org/documentation/#link-list-of-built-in-validation-methods), should be listed below in additional-methods-list.

Arkni:
@nebrot 
Good catch, thanks for pointing out :)

PR #1614, opened 2015-10-13
Fixed #1523

### LINKED ISSUE #1523: "aria-describedby" being filled with repeating values. Cause: "errorID" being stored as escaped string
In /src/core.js starting at line 799 in version 1.14.0

It is storing the string errorID in the attribute as an escaped sting, so it will never match the Regex later when it is queried again.

Here is the code in question:

```javascript
                    errorID = error.attr( "id" ).replace( /(:|\.|\[|\]|\$)/g, "\\$1");
                    // Respect existing non-error aria-describedby
                    if ( !describedBy ) {
                        describedBy = errorID;
                    } else if ( !describedBy.match( new RegExp( "\\b" + errorID + "\\b" ) ) ) {
                        // Add to end of list if not already present
                        describedBy += " " + errorID;
                    }
                    $( element ).attr( "aria-describedby", describedBy );
```
		    

Here is my suggested code fix:

```javascript
                    errorID = error.attr( "id" );
                    var escaped_errorID = errorID.replace( /(:|\.|\[|\]|\$)/g, "\\$1");
                    // Respect existing non-error aria-describedby
                    if ( !describedBy ) {
                        describedBy = errorID;
                    } else if ( !describedBy.match( new RegExp( "\\b" + escaped_errorID + "\\b" ) ) ) {
                        // Add to end of list if not already present
                        describedBy += " " + errorID;
                    }
                    $( element ).attr( "aria-describedby", describedBy );
```

Arkni:
As suggested by Jörn in [#1523 (comment)](https://github.com/jzaefferer/jquery-validation/issues/1523#issuecomment-122916989):
>Could you look into writing a unit test that captures what you have in your jsfiddle? Once that works, the suggested fix is much easier to verify and the test will prevent further regressions.

This pull request is a continuation of #1587 (so you can close it in favor of this) :
>The error ID was being escaped for use in a regex, but these escaped characters were also being passed on to the aria-describedby value set in the HTML, which means the ID doesn't match up with the ID of the element it's referring to.

**Note:**
The first commit contains the unit test, while the second one contains the fix.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -772,6 +772,7 @@ $.extend( $.validator, {
 				error = this.errorsFor( element ),
 				elementID = this.idOrName( element ),
 				describedBy = $( element ).attr( "aria-describedby" );
+
 			if ( error.length ) {
 				// refresh error/success class
 				error.removeClass( this.settings.validClass ).addClass( this.settings.errorClass );
@@ -803,15 +804,15 @@ $.extend( $.validator, {
 				if ( error.is( "label" ) ) {
 					// If the error is a label, then associate using 'for'
 					error.attr( "for", elementID );
-				} else if ( error.parents( "label[for='" + elementID.replace( /'/g, "\\'" ) + "']" ).length === 0 ) {
+				} else if ( error.parents( "label[for='" + this.escapeCssMeta( elementID ) + "']" ).length === 0 ) {
 					// If the element is not a child of an associated label, then it's necessary
 					// to explicitly apply aria-describedby
 
-					errorID = error.attr( "id" ).replace( /(:|\.|\[|\]|\$)/g, "\\$1" );
+					errorID = error.attr( "id" );
 					// Respect existing non-error aria-describedby
 					if ( !describedBy ) {
 						describedBy = errorID;
-					} else if ( !describedBy.match( new RegExp( "\\b" + errorID + "\\b" ) ) ) {
+					} else if ( !describedBy.match( new RegExp( "\\b" + this.escapeCssMeta( errorID ) + "\\b" ) ) ) {
 						// Add to end of list if not already present
 						describedBy += " " + errorID;
 					}
@@ -822,7 +823,7 @@ $.extend( $.validator, {
 					if ( group ) {
 						$.each( this.groups, function( name, testgroup ) {
 							if ( testgroup === group ) {
-								$( "[name='" + name.replace( /'/g, "\\'" ) + "']", this.currentForm )
+								$( "[name='" + this.escapeCssMeta( name ) + "']", this.currentForm )
 									.attr( "aria-describedby", error.attr( "id" ) );
 							}
 						} );
@@ -841,20 +842,26 @@ $.extend( $.validator, {
 		},
 
 		errorsFor: function( element ) {
-			var name = this.idOrName( element ).replace( /'/g, "\\'" ),
+			var name = this.escapeCssMeta( this.idOrName( element ) ),
 				describer = $( element ).attr( "aria-describedby" ),
 				selector = "label[for='" + name + "'], label[for='" + name + "'] *";
 
 			// aria-describedby should directly reference the error element
 			if ( describer ) {
-				selector = selector + ", #" + describer
-					.replace( /'/g, "\\'" )
+				selector = selector + ", #" + this.escapeCssMeta( describer )
 					.replace( /\s+/g, ", #" );
 			}
 
 			return this
 				.errors()
 				.filter( selector );
+		},
+
+		// See https://api.jquery.com/category/selectors/, for CSS
+		// meta-characters that should be escaped in order to be used with JQuery
+		// as a literal part of a name/id or any selector.
+		escapeCssMeta: function( string ) {
+			return string.replace( /(!|"|#|\$|%|&|'|\(|\)|\*|\+|,|\.|\/|:|;|<|=|>|\?|@|\[|\\|\]|\^|`|\{|\||\}|~)/g, "\\$1");
 		},
 
 		idOrName: function( element ) {
@@ -877,7 +884,7 @@ $.extend( $.validator, {
 		},
 
 		findByName: function( name ) {
-			return $( this.currentForm ).find( "[name='" + name.replace( /'/g, "\\'" ) + "']" );
+			return $( this.currentForm ).find( "[name='" + this.escapeCssMeta( name ) + "']" );
 		},
 
 		getLength: function( value, element ) {
--- a/test/error-placement.js
+++ b/test/error-placement.js
@@ -291,6 +291,42 @@ test( "test existing non-label used as error element", function(assert) {
 	assert.noErrorFor( field );
 });
 
+test( "test aria-describedby with input names contains special characters (:, ., [, ], $)", function() {
+	var form = $( "#testForm21" ),
+		field = $( "#testForm21\\[\\]\\.value" );
+
+	equal( field.attr( "aria-describedby" ), undefined );
+
+	form.validate( {
+		errorElement: "span",
+		errorPlacement: function() {
+			// Do something
+		}
+	} );
+
+	// Validate the element
+	ok( !field.valid() );
+	equal( field.attr( "aria-describedby" ), "testForm21[].value-error" );
+
+	// Re-run validation
+	field.val( "some" );
+	field.trigger( "keyup" );
+
+	field.val( "something" );
+	field.trigger( "keyup" );
+
+	// `aria-describedby` should remain the same as before.
+	equal( field.attr( "aria-describedby" ), "testForm21[].value-error" );
+
+	// Re-run validation
+	field.val( "something something" );
+	field.trigger( "keyup" );
+
+	ok( field.valid() );
+	// `aria-describedby` should remain the same as before.
+	equal( field.attr( "aria-describedby" ), "testForm21[].value-error" );
+});
+
 test( "test existing non-error aria-describedby", function( assert ) {
 	expect( 8 );
 	var form = $( "#testForm17" ),
--- a/test/error-placement.js
+++ b/test/error-placement.js
@@ -291,9 +291,9 @@ test( "test existing non-label used as error element", function(assert) {
 	assert.noErrorFor( field );
 });
 
-test( "test aria-describedby with input names contains special characters (:, ., [, ], $)", function() {
+test( "test aria-describedby with input names contains CSS-selector meta-characters", function() {
 	var form = $( "#testForm21" ),
-		field = $( "#testForm21\\[\\]\\.value" );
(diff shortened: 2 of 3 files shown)


Arkni:
jsfiddle with the fix: http://jsfiddle.net/Arkni/mk4p9fLf/

src/core.js
```diff
@@ -848,7 +848,7 @@ $.extend( $.validator, {
 			// aria-describedby should directly reference the error element
 			if ( describer ) {
 				selector = selector + ", #" + describer
-					.replace( /'/g, "\\'" )
+					.replace( /('|:|\.|\[|\]|\$)/g, "\\$1" )
```
staabm: Would it make sense to extract this pattern into a var so this and line 814 cannot get out of sync?

Arkni: Line 814 is in `showLabel` method, and line 851 is in `errorFor` method. So what do you suggest to solve this ? a global variable or something else ?

staabm: I meant line https://github.com/jzaefferer/jquery-validation/pull/1614/files#diff-7eb52b366866677666470e019283c8eaR814

We can just use a local var, dont need a global one.

Arkni: Know I understand what you mean.

Will do thanks :)

test/error-placement.js
```diff
@@ -291,6 +291,42 @@ test( "test existing non-label used as error element", function(assert) {
 	assert.noErrorFor( field );
 });
 
+test( "test aria-describedby with input names contains special characters (:, ., [, ], $)", function() {
```
staabm: are there even more selector-meta-characters which have this problem? e.g. `(`, `)`

staabm: please note in the test description that we are talking about css-selector meta characters contained in fieldnames

test/error-placement.js
```diff
@@ -291,6 +291,42 @@ test( "test existing non-label used as error element", function(assert) {
...
+	equal( field.attr( "aria-describedby" ), "testForm21[].value-error" );
+
+	// Re-run validation
+	field.val( "some" );
+	field.trigger( "keyup" );
+
+	field.val( "something" );
+	field.trigger( "keyup" );
+
+	// `aria-describedby` should remain the same as before.
```
staabm: please put this message into `equal` as a 3rd arg

test/error-placement.js
```diff
@@ -291,6 +291,42 @@ test( "test existing non-label used as error element", function(assert) {
...
+	equal( field.attr( "aria-describedby" ), "testForm21[].value-error" );
+
+	// Re-run validation
+	field.val( "something something" );
+	field.trigger( "keyup" );
+
+	ok( field.valid() );
+	// `aria-describedby` should remain the same as before.
```
staabm: same here: 3rd arg

test/error-placement.js
staabm: I guess we could get some inspiration here: https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape

Arkni: I think all of these meta-characters will result in the problem:
```
!"#$%&'()*+,./:;<=>?@[\]^`{|}~
```

mentioned here: https://api.jquery.com/category/selectors/

test/error-placement.js
Arkni: Will fix that :)

test/error-placement.js
Arkni: See also: https://mathiasbynens.be/notes/css-escapes

staabm: Great, lets support everything jQuery supports (so the charlist you mentioned above). I dont think we need to be 100% compilant with all those unicode special chars etc.

staabm:
@Arkni let me just thank you again for the awesome job you are doing!

test/error-placement.js
Arkni: Working on that right now :)

Arkni:
> @Arkni let me just thank you again for the awesome job you are doing!

It's all my pleasure :)

Arkni:
@staabm 

Do I need to extend the unit test to cover at least 2 or 3 or even 4 meta-characters mentioned above ?

staabm:
@Arkni I would put all of them into one name

Arkni:
> @Arkni I would put all of them into one name

Best option :+1:

Arkni:
@staabm 
Done!

src/core.js
```diff
@@ -848,7 +853,8 @@ $.extend( $.validator, {
 			// aria-describedby should directly reference the error element
 			if ( describer ) {
 				selector = selector + ", #" + describer
-					.replace( /'/g, "\\'" )
+					// See https://api.jquery.com/category/selectors/
+					.replace( /(!|"|#|\$|%|&|'|\(|\)|\*|\+|,|\.|\/|:|;|<|=|>|\?|@|\[|\\|\]|\^|`|\{|\||\}|~)/g, "\\$1" )
```
staabm: We want to re-use `theregex` write?

src/core.js
```diff
@@ -771,7 +771,12 @@ $.extend( $.validator, {
 			var place, group, errorID,
 				error = this.errorsFor( element ),
 				elementID = this.idOrName( element ),
-				describedBy = $( element ).attr( "aria-describedby" );
+				describedBy = $( element ).attr( "aria-describedby" ),
+				// See https://api.jquery.com/category/selectors/, for CSS
+				// meta-characters that should be escaped in order to be used with JQuery
+				// as a literal part of a name/id or any selector.
+				theregex = /(!|"|#|\$|%|&|'|\(|\)|\*|\+|,|\.|\/|:|;|<|=|>|\?|@|\[|\\|\]|\^|`|\{|\||\}|~)/g;
```
staabm: Rename to `rCssMeta`

test/error-placement.js
```diff
@@ -291,6 +291,40 @@ test( "test existing non-label used as error element", function(assert) {
 	assert.noErrorFor( field );
 });
 
+test( "test aria-describedby with input names contains CSS-selector meta-characters", function() {
+	var form = $( "#testForm21" ),
+		field = $( "#testForm21\\!\\#\\$\\%\\&\\\'\\(\\)\\*\\+\\,\\.\\/\\:\\;\\<\\=\\>\\?\\@\\[\\\\\\]\\^\\`\\{\\|\\}\\~" );
```
staabm: One backslash too much after `&`?

src/core.js
Arkni: You mean something like this:
```js
this.rCssMeta = /.../;
```

Right ?

staabm: Just like above `..replace(rCssMeta, "\\$1)`

Arkni: The regex is used in two different methods. I think using it like this will do the trick (I'm not sure)
```js
this.rCssMeta = /.../;
```

test/error-placement.js
Arkni: make sense only if you wrap the whole sting in single quotes (That's why I escaped it like that, but forgot to remove it when wrapping inside double quotes).

src/core.js
staabm: Ah, now I get the problem. A pity we can use a closure in both methods. 

Maybe its easiesr to create a private method which returns the regex or even better one which escapes a given string?

Arkni:
@staabm 

Done 8327818

staabm:
Perfect, thx

Arkni:
You're welcome :)

src/core.js
```diff
@@ -822,7 +823,7 @@ $.extend( $.validator, {
 					if ( group ) {
 						$.each( this.groups, function( name, testgroup ) {
 							if ( testgroup === group ) {
-								$( "[name='" + name.replace( /'/g, "\\'" ) + "']", this.currentForm )
+								$( "[name='" + this.escapeCssMeta( name ) + "']", this.currentForm )
```
staabm: @Arkni I just did some manual testing in a real world app with todays master version. I got errors like

```
TypeError: this.escapeCssMeta is not a function

.prototype.showLabel/<()     jquery.validate.js:927
.each()   jquery.js:2
.prototype.showLabel()    jquery.validate.js:925
.prototype.defaultShowErrors()    jquery.validate.js:833
.prototype.showErrors()   jquery.validate.js:499
.prototype.element()    jquery.validate.js:468
.onfocusout()
```

with a `onfocusout` default setting like

```js
$.validator.setDefaults({
// ...snip
        onfocusout: function(element) {
        	// activate immediate-as-you-type validation, by overriding default behaviour
            var isValid = this.element(element);
            
        	// delete the wrapping "field-with-errors" element, when fields get valid
            if (isValid) {
            	var $el = $(element);
            	if ($el.parent().is(".field-with-errors")) {
            		$el.unwrap();
            	}
            }
        },
// snap...
```

atm I dont have time to investigate it furhter. just to let you know in the meantime.

Arkni: I think It's my fault, as I forgot that the `this` inside `$.each( ... )` refer to the current value of the `object|array` we iterate over.

I'll investigate the issue and put together a PR to fix it with a unit test ASAP

Arkni: Hi @staabm 
The problem was as I said in my previous comment, `this` inside `$.each( ... )` refers to the current value of the `object|array` we iterate over, not the plugin instance. So this patch should fix it:

```diff
diff --git a/src/core.js b/src/core.js
index 7787a8e..a3d37e9 100644
--- a/src/core.js
+++ b/src/core.js
@@ -843,7 +843,7 @@ $.extend( $.validator, {
 		},
 
 		showLabel: function( element, message ) {
-			var place, group, errorID,
+			var place, group, errorID, v,
 				error = this.errorsFor( element ),
 				elementID = this.idOrName( element ),
 				describedBy = $( element ).attr( "aria-describedby" );
@@ -904,9 +904,10 @@ $.extend( $.validator, {
 					// If this element is grouped, then assign to all elements in the same group
 					group = this.groups[ element.name ];
 					if ( group ) {
-						$.each( this.groups, function( name, testgroup ) {
+						v = this;
+						$.each( v.groups, function( name, testgroup ) {
 							if ( testgroup === group ) {
-								$( "[name='" + this.escapeCssMeta( name ) + "']", this.currentForm )
+								$( "[name='" + v.escapeCssMeta( name ) + "']", v.currentForm )
 									.attr( "aria-describedby", error.attr( "id" ) );
 							}
 						} );
```

Right now, I don't have the time to create a PR with a unit test. But will do soon.

About the test, what can we test to check that this fix work ?

staabm: It sounds like we dont have a test covering the `.groups` feature

Arkni: See:
* https://github.com/jzaefferer/jquery-validation/blob/master/test/test.js#L1018-L1033
* https://github.com/jzaefferer/jquery-validation/blob/master/test/messages.js#L12-L43

PR #1621, opened 2015-10-23
Build: Adds CommonJS-support in build process. Enhances PR #1597

codeclown:
See https://github.com/jzaefferer/jquery-validation/pull/1597

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -25,7 +25,9 @@ banner = "/*!\n" +
 umdStart = "(function( factory ) {\n" +
 	"\tif ( typeof define === \"function\" && define.amd ) {\n";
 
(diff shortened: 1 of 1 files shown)


staabm:
@codeclown did you test this change? I am asking because I am not at a workstation right now and we dont have test coverage for this kind of things

codeclown:
Yes, tested locally. Unfortunately the dist-version is not included in the repo because it requires integration, but I have included it [here](https://gist.github.com/codeclown/45aeaf3f10f714c274e9) so anyone can test requiring it as well.

staabm:
@Arkni any outstanding issues?

Arkni:
LGTM :)

PR #1405, opened 2015-02-18
HTML5 step attribute support

etylsarin:
lightweight update to support step attribute for numeric values.
Fixes issue #1295

PR #1638, opened 2015-11-07
Core: HTML5 step attribute support

### LINKED ISSUE #1295: Step attribute is not supported
Hello and thanks for your great tool,

The "step" attribute in an input type "number" is not correctly supported.

Exemple
```
<input type="number" step="0.01">
```
Without your plugin, in chrome 38, while submitting the form, the browser show an error for the value 15.511 (no error for 15.51)

With your plugin, no error is thrown.

Is it normal that your plugin completely avoid browser's validations ?

Thanks

etylsarin:
This PR adds support of step attribute for numeric values.
It follows-up on #1405 which was closed due to huge number of commits.
Documentation PR: jzaefferer/validation-content#17
Fixes #1295


src/core.js
```diff
@@ -1316,6 +1317,14 @@ $.extend( $.validator, {
 			return this.optional( element ) || ( value >= param[ 0 ] && value <= param[ 1 ] );
 		},
 
+		// http://jqueryvalidation.org/step-method/
+		step: function( value, element, param ) {
+
+			// Works only for number and range input types
```
staabm: What do you think about throwing a exception, or at least log a error into the console when in debug mode, in case someone uses this feature with unsupported input types?

etylsarin: @staabm check it out mate. Really hope we can close it finally.

staabm:
manually merged with b3c07e39571a8e6a0583ef9a50ca10681374890e. I had to fix the unit tests for Firefox in a followup commit though.

@etylsarin thank you.

PR #1639, opened 2015-11-10
Changed email validation so emails require domain.

acodeinprogress:


acodeinprogress:
The tests show that an email such as name@domain is valid. Is this the way it should be?
Thanks!

staabm:
See https://github.com/jzaefferer/jquery-validation/blob/master/CONTRIBUTING.md

> IMPORTANT NOTE ABOUT EMAIL VALIDATION. As of version 1.12.0 this plugin is using the same regular expression that the HTML5 specification suggests for browsers to use. We will follow their lead and use the same check. If you think the specification is wrong, please report the issue to them. If you have different requirements, consider using a custom method.

see https://github.com/jzaefferer/validation-content/pull/14/files for how to use a custom email pattern

acodeinprogress:
So sorry! Lesson learned. Thank you!

staabm:
@acodeinprogress no problem, feel free to participate in any other form :-).

Email format is like a running gag though, since we get ~1 report per week ... Seems that people do not read the readme or search the issue tracker.

DePalmo:
Then why not making this more visible and offer a solution that would work for both? I came here because exactly the same reason. Your validation plugin (is quite great!) says email is valid, but my PHP code says it's not. And I never heard that name@domain could be a valid email address. Maybe in closed local network, but not on the internet.

staabm:
@DePalmo where do we miss to inform the user about the email validation pattern?

AFAIR we have it in the README & CONTRIBUTION File on github, as well as the official docs for the email method.

DePalmo:
I'm only saying it because:
1. I did look at the docs (but not readme) and didn't found this mentioned
2. you mentioned that you have an issue opened about this almost each week

So, you could add this note to the documentation available on plugin's website. That would actually make it visible.

staabm:
See http://jqueryvalidation.org/email-method/ its mentioned and even linked in the docs.

Thx for suggesting ways to improve, though

DePalmo:
Then I must be blind. On the link provided, I only see that we can use a custom validation for email but there's no notice that emails are validated strictly by HTML5 specs and that most coders most probably won't like it.

It would be nice if there would be a notice about this for us, "dumb people". I'm also sure it will bring down the "issues reported" because we will know what's wrong and will look for a solution from there on and not came here reporting issues that are not there. Bonus would be to have a working example for custom email rule on your website.

And no problem, just trying to help.

staabm:
Ah now I get what you think is missing. Feel free to provide a PR at https://github.com/jzaefferer/validation-content

Thx

PR #1641, opened 2015-11-11
Fixes #1632: Error hidden but input error class not removed

### LINKED ISSUE #1632: Error hidden, but input error class not removed
https://jsfiddle.net/p9m1u8ry/

Any easy way to fix it?

Thanks

Arkni:
When a field has one rule with option `depends` specified and this
dependency mismatch, the plugin will remove it, so that field will not be
validated on submit by the the rule in question. So if that field is
already invalid, only the error message will be hidden and the error class
will remain.

Fixes #1632

(The commit message may need a reword)

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -489,9 +489,15 @@ $.extend( $.validator, {
 			this.submitted = {};
 			this.prepareForm();
 			this.hideErrors();
-			var i, elements = this.elements()
+			var elements = this.elements()
 				.removeData( "previousValue" )
 				.removeAttr( "aria-invalid" );
+
+			this.resetElements( elements );
+		},
+
+		resetElements: function( elements ) {
+			var i;
 
 			if ( this.settings.unhighlight ) {
 				for ( i = 0; elements[ i ]; i++ ) {
@@ -1190,6 +1196,7 @@ $.extend( $.validator, {
 				if ( keepRule ) {
 					rules[ prop ] = val.param !== undefined ? val.param : true;
 				} else {
+					$.data( element.form, "validator" ).resetElements( $( element ) );
 					delete rules[ prop ];
 				}
 			}
--- a/test/error-placement.js
+++ b/test/error-placement.js
@@ -407,3 +407,36 @@ test( "test id/name containing single quotes", function() {
 	equal( v.invalidElements()[ 1 ], checkboxElement[ 0 ], "The text element should be invalid" );
 	equal( v.invalidElements()[ 2 ], radioElement[ 0 ], "The text element should be invalid" );
 } );
+
+test( "#1632: Error hidden, but input error class not removed", function() {
+	var v = $( "#testForm23" ).validate( {
+			rules: {
+				box1: {
+					required: {
+						depends: function() {
+							return !!$( "#box2" ).val();
+						}
+					}
+				},
+				box2: {
+					required: {
+						depends: function() {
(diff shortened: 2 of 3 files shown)


Arkni:
@staabm 
Could you take a look at this fix ? is it right doing it that way ?

src/core.js
```diff
@@ -482,14 +482,17 @@ $.extend( $.validator, {
...
 
+			this.resetElements( elements );
+		},
+
+		resetElements: function( elements ) {
 			if ( this.settings.unhighlight ) {
-				for ( i = 0; elements[ i ]; i++ ) {
-					this.settings.unhighlight.call( this, elements[ i ],
-						this.settings.errorClass, "" );
+				for ( var i = 0; elements[ i ]; i++ ) {
```
staabm: Var declarations are only allowed at the start of the method (per CS)

staabm:
Need to think about it when time allows

src/core.js
Arkni: My bad, forgot about that.
will fix that.

staabm: Any chance to enforce this rule with our stylechecker?

Arkni: Yes, there is a JSCS rule that disallow that: [requireVarDeclFirst](http://jscs.info/rule/requireVarDeclFirst)

But using it will require fixing a lot of issues in test and src files.

What do you think ?

staabm: After we landed all the complete PRs we could do that in a following step. Not important right now.

Arkni:
> Need to think about it when time allows

Take your time :)

src/core.js
Arkni: OK

staabm:
I still not had the time to look at it, but maybe @Arkni you could rebase it in the meantime, as it no longer merges cleanly because of other PRs beeing merged.

Arkni:
> I still not had the time to look at it

No rush :)

> but maybe @Arkni you could rebase it in the meantime, as it no longer merges cleanly because of other PRs beeing merged.

Done!

Arkni:
Currently, I'm not seeing this fix as the proper one, it needs some analysis to see in which place in the source code the element should be reset.

So, let this PR open until we have a better visual of how to fix it ;)

Arkni:
@staabm 
Does this PR need any adjustment I should do?

src/core.js
```diff
@@ -489,10 +489,16 @@ $.extend( $.validator, {
...
 			this.hideErrors();
-			var i, elements = this.elements()
+			var elements = this.elements()
 				.removeData( "previousValue" )
 				.removeAttr( "aria-invalid" );
 
+			this.resetElements( elements );
+		},
+
+		resetElements: function( elements ) {
```
staabm: I guess we should move this method somewhere that it is not part of the published api.

Arkni: I guess not, as there are a lot of methods added to prototype (which means published) but not documented anywhere in the source code nor the documentation.

But if there is a better place for this method, please let me know.

staabm: yeah, after re-reading the source I agree that we used to doc public api and leave the rest in the prototype. so we can leave it as is.

staabm:
landed in ef6821aba333f823449730d59b80155f788ade8f. thx @Arkni good job.

Arkni:
You're welcome :)

PR #1642, opened 2015-11-12
Added missing pattern validator message in Polish

adamwojtkiewicz:


staabm:
please squash your commits into one and prefix the commit message with `Localisation`

PR #1643, opened 2015-11-12
Core: fixed typo in internal findLastActive method

staabm:
`lastActive` was checked twice within the same expression. fix that.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -562,7 +562,7 @@ $.extend( $.validator, {
 			var lastActive = this.lastActive;
 			return lastActive && $.grep( this.errorList, function( n ) {
 				return n.element.name === lastActive.name;
-			} ).length === 1 && lastActive;
+			} ).length === 1;
 		},
 
 		elements: function() {


staabm:
@Arkni please double check

src/core.js
```diff
@@ -562,7 +562,7 @@ $.extend( $.validator, {
 			var lastActive = this.lastActive;
 			return lastActive && $.grep( this.errorList, function( n ) {
 				return n.element.name === lastActive.name;
-			} ).length === 1 && lastActive;
```
Arkni: I think there is no typo here.
The first check test if `lastActive` is not null or undefined and then check for the second condition and if the second pass, the method will return `lastActive`.

So this method return either `false` or `lastActive` (`Element`)

staabm: hmm tests are failling now, so my change did break things.

I still dont get where `lastActive` is changed between line 563 and 565 so, I dont get why it is checked against to be "truethy" again...

Arkni: > I still dont get where lastActive is changed between line 563 and 565 so, I dont get why it is checked against to be "truethy" again...

No, you get it wrong. The expression it's not clear and need a refractoring, something like this:
```js
		findLastActive: function() {
			var lastActive = this.lastActive,
				arr;

			if ( lastActive ) {
				arr = $.grep( this.errorList, function( n ) {
					return n.element.name === lastActive.name;
				} );

				return arr.length === 1 ? lastActive : false;
			}

			return false;
		},
```

Arkni: > hm this would be true if we had a || operator, but we have &&, no?

Where ? the last one ?
If yes, let's me explain by an example:
```
true && "Something"
> "Something"

false && "Something"
> false
```

But
```
true || "Something"
> true

false || "Something"
> "Something"
```

The second case is not the behavior the one who implemented the `findLastActive` method attends to get from it.

staabm: thanks for your examples... now I get whats going on and I agree we need to refactor things, because it is misleading, at least for non-js devs ;-).

But hey, we had unit tests which prevented that I am doing stupid things... ;-)

Arkni: You're welcome :)

> I agree we need to refactor things, because it is misleading

We have to look for a better refractoring (better than the one I provided above).

> But hey, we had unit tests which prevented that I am doing stupid things... ;-)

That's just the effect of JavaScript's trickiness :D

PR #1644, opened 2015-11-12
Core: move message processing from `formatAndAdd` to `defaultMessage`

### LINKED ISSUE #741: Move message processing from formatAndAdd to defaultMessage
With remote validation, messages aren't processed through `formatAndAdd`. Since that method does more than what `defaultMessage` does, something like "can't use {0}" fails when used with the remote method.

Arkni:
Fixes #741

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -213,6 +213,9 @@ $.validator.format = function( source, params ) {
 			return $.validator.format.apply( this, args );
 		};
 	}
+	if ( params === undefined ) {
+		return source;
+	}
 	if ( arguments.length > 2 && params.constructor !== Array  ) {
 		params = $.makeArray( arguments ).slice( 1 );
 	}
@@ -739,26 +742,29 @@ $.extend( $.validator, {
 			return undefined;
 		},
 
-		defaultMessage: function( element, method ) {
-			return this.findDefined(
-				this.customMessage( element.name, method ),
-				this.customDataMessage( element, method ),
+		defaultMessage: function( element, rule ) {
+			var message = this.findDefined(
+					this.customMessage( element.name, rule.method ),
+					this.customDataMessage( element, rule.method ),
 
-				// 'title' is never undefined, so handle empty string as undefined
-				!this.settings.ignoreTitle && element.title || undefined,
-				$.validator.messages[ method ],
-				"<strong>Warning: No message defined for " + element.name + "</strong>"
-			);
-		},
-
-		formatAndAdd: function( element, rule ) {
-			var message = this.defaultMessage( element, rule.method ),
+					// 'title' is never undefined, so handle empty string as undefined
+					!this.settings.ignoreTitle && element.title || undefined,
+					$.validator.messages[ rule.method ],
+					"<strong>Warning: No message defined for " + element.name + "</strong>"
+				),
 				theregex = /\$?\{(\d+)\}/g;
 			if ( typeof message === "function" ) {
 				message = message.call( this, rule.parameters, element );
 			} else if ( theregex.test( message ) ) {
 				message = $.validator.format( message.replace( theregex, "{$1}" ), rule.parameters );
 			}
+
+			return message;
+		},
+
+		formatAndAdd: function( element, rule ) {
+			var message = this.defaultMessage( element, rule );
+
 			this.errorList.push( {
 				message: message,
 				element: element,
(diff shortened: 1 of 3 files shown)


src/core.js
```diff
@@ -213,6 +213,9 @@ $.validator.format = function( source, params ) {
 			return $.validator.format.apply( this, args );
 		};
 	}
+	if ( params === undefined ) {
```
staabm: Unrelated change?

Arkni: I did it on purpose because of this line [1013](https://github.com/jzaefferer/jquery-validation/pull/1644/files#diff-7eb52b366866677666470e019283c8eaR1013) in this PR.

Should I omit it and change the line 1013 with the following:
```js
message: this.defaultMessage( element, { method: "remote", parameters: "" } )
```

staabm:
thanks!

src/core.js
```diff
@@ -739,26 +742,29 @@ $.extend( $.validator, {
...
-				this.customMessage( element.name, method ),
-				this.customDataMessage( element, method ),
-
-				// 'title' is never undefined, so handle empty string as undefined
-				!this.settings.ignoreTitle && element.title || undefined,
-				$.validator.messages[ method ],
-				"<strong>Warning: No message defined for " + element.name + "</strong>"
-			);
-		},
+		defaultMessage: function( element, rule ) {
```
Maks3w: This is a BC Break. Previusly the second argument was an string and now its an object.

Maks3w:
```js
if (rule typeof string) {
  rule = {method: rule}
}
```

Arkni: You're absolutely right!
But just a note, undocumented methods are, in the first place, used internally, and only a bunch o methods are intended to be used elsewhere:

>The validate method returns a Validator object that has a few public methods that you can use to trigger validation programmatically or change the contents of the form. The validator object has more methods, but only those documented here are intended for usage.
>
* Validator.form() – Validates the form.
* Validator.element() – Validates a single element.
* Validator.resetForm() – Resets the controlled form.
* Validator.showErrors() – Show the specified messages.
* Validator.numberOfInvalids() – Returns the number of invalid fields.
>
There are a few static methods on the validator object:
>
* jQuery.validator.addMethod() – Add a custom validation method.
* jQuery.validator.format() – Replaces {n} placeholders with arguments.
* jQuery.validator.setDefaults() – Modify default settings for validation.
* jQuery.validator.addClassRules() – Add a compound class method.
>
...

But I can fix that in a followup proposal to maintain backward compatibility with code that still relaying on the old behavior.

Thanks for the note!

Maks3w: How can we write custom validation rules then?

staabm: Like any of the additional methos are written
https://github.com/jzaefferer/jquery-validation/tree/master/src/additional

Maks3w: None of them use a remote validation.

Arkni: I'll create a PR soon with the following patch:
```diff
 src/core.js  | 8 +++++++-
 test/test.js | 2 +-
 2 files changed, 8 insertions(+), 2 deletions(-)

diff --git a/src/core.js b/src/core.js
index 743f028..b497c4f 100644
--- a/src/core.js
+++ b/src/core.js
@@ -803,6 +803,10 @@ $.extend( $.validator, {
 		},
 
 		defaultMessage: function( element, rule ) {
+			if ( typeof rule === "string" ) {
+				rule = { method: rule };
+			}
+
 			var message = this.findDefined(
 					this.customMessage( element.name, rule.method ),
 					this.customDataMessage( element, rule.method ),
@@ -1067,10 +1071,12 @@ $.extend( $.validator, {
 		},
 
 		previousValue: function( element, method ) {
+			method = typeof method === "string" && method || "remote";
+
 			return $.data( element, "previousValue" ) || $.data( element, "previousValue", {
 				old: null,
 				valid: true,
-				message: this.defaultMessage( element, { method: method } )
+				message: this.defaultMessage( element, method )
 			} );
 		},
 
diff --git a/test/test.js b/test/test.js
index 12541c2..05ec156 100644
--- a/test/test.js
+++ b/test/test.js
@@ -678,7 +678,7 @@ test( "option: errorClass with multiple classes", function() {
 
 test( "defaultMessage(), empty title is ignored", function() {
 	var v = $( "#userForm" ).validate();
-	equal( v.defaultMessage( $( "#username" )[ 0 ], { method: "required", parameters: true } ), "This field is required." );
+	equal( v.defaultMessage( $( "#username" )[ 0 ], "required" ), "This field is required." );
 } );
 
 test( "#741: move message processing from formatAndAdd to defaultMessage", function() {
```

If there are other things I should take care off, please mention them and I will update before sending the PR

Maks3w: I don't see why change `message: this.defaultMessage( element, { method: method } )`

May add some kind of deprecation comment about the old parameter type so it could be removed on next major version

Arkni: Make sense.
I'll do the update soon.
Thanks!

PR #1646, opened 2015-11-13
Core: errorList should contain only the errors that it should

### LINKED ISSUE #1618: Errorlist containing more errors than it should
Hi there,

During development I came across an issue with jQuery Validation plugin.

It seems that the ErrorList isn't cleared when the first input is ignored causing pollution.

Here is a JSFiddle of the issue. https://jsfiddle.net/xLcz7ts9/
Open the Developer console and watch the console log.

You'll see that the ErrorList slowly grows larger and larger.
When this occurs on a page with lots of inputs just 4 iterations is all it takes to freeze the web browser.

My workaround (which may be used as the solution) is to change the `element: function (element)` from the following:
```js
            element: function (element) {
                ...
                if (checkElement === undefined) {
                    delete this.invalid[cleanElement.name];
                } else {
                    this.prepareElement(checkElement);
                    this.currentElements = $(checkElement);
                ...
            },
```

To the following:
```js
            element: function (element) {
                ...
                if (checkElement === undefined) {
                    this.prepareElement(cleanElement);
                    this.currentElements = $(cleanElement);
                    delete this.invalid[cleanElement.name];
                } else {
                    this.prepareElement(checkElement);
                    this.currentElements = $(checkElement);
                ...
            },
```

P.S. I'm using the lastest code from the master branch, this was tested in Chrome and IE11, and should be reproducible in any browser.

Arkni:
When using the `valid` method on a list of inputs and there is an ignored
input in that list, the errorList is not cleared causing it to slowly
growing larger and larger when calling `valid` several times.

Fixes #1618

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -106,7 +106,9 @@ $.extend( $.fn, {
 			validator = $( this[ 0 ].form ).validate();
 			this.each( function() {
 				valid = validator.element( this ) && valid;
-				errorList = errorList.concat( validator.errorList );
+				if ( !valid ) {
+					errorList = errorList.concat( validator.errorList );
+				}
 			} );
 			validator.errorList = errorList;
 		}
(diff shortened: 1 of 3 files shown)


staabm:
merged manually via 43ffa811801c2e3ae32a0fac739e9c22d9470f2b, thanks @Arkni great job!

Arkni:
You're welcome :)

PR #1647, opened 2015-11-13
localization: adding "tel" and "pattern" string to "zh"

horans:
(added in the beginning of all strings.)

### DIFF
--- a/src/localization/messages_zh.js
+++ b/src/localization/messages_zh.js
@@ -3,6 +3,8 @@
  * Locale: ZH (Chinese, 中文 (Zhōngwén), 汉语, 漢語)
  */
 $.extend( $.validator.messages, {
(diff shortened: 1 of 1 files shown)


staabm:
Please prepend the commit message with "Localisation:"

horans:
these 2 strings is not included in other language files, so i am not sure this commit should be categorized as "localization".
anyway, edited.

horans:
these 2 strings is not included in other language files, so i am not sure this commit should be categorized as "localization".
anyway, edited.

staabm:
Wha do you think we need those strings when they are not part of other lang files?

horans:
these two strings refer to the input's html5 feature. i used them in my project, at least.
in my opinion, all language file should include them, especially for the "pattern" string.

Issue #659, opened 2013-02-27
Try to remove name attribute requirement


jzaefferer:
I can't tell anymore why the name attribute requirement is there in the first place. Need to figure that out or just drop it, with whatever consequences that has.

Motivation:
- Rules and messages can be configured inline, so no need to reference elements by name
- Having multiple elements with the same name comes up every once in a while, ala `address[]`. Explicit indices usually work (`address[0]`, `address[1]` etc), but are less convenient.

See also #654 for a bad workaround.

jzaefferer:
I'm sorry for the lack of activity on this issue. Instead of leaving it open any longer, I decided to close old issues without trying to address them, to longer give the false impression that it will get addressed eventually.

To the reporter (or anyone else interested in this issue): If you're affected by the same issue, consider opening a new issue, with a testpage that shows the issue. Even better, try to fix the issue yourself, and improve the project by sending a pull request. This may seem daunting at first, but you'll likely learn some useful skills that you can apply elsewhere as well. And you can help keep this project alive. [We've documented how to do these things, too](https://github.com/jzaefferer/jquery-validation/blob/master/CONTRIBUTING.md). A patch is worth a thousand issues!

MisinformedDNA:
Wow. So you want to give the false impression that you are actually solving problems? Sad.

This issue has been brought up so many times as seen above and the repro is clear and you just close it out again. It just further fragments the discussion of the issue.

chrishaines:
Perhaps this could be solved by supporting an optional html5 data attribute (maybe ```data-validator-name```).  It could then fallback to the name if it does not exist. Instead of referencing element.name directly, it could use a method like so.

```javascript
getName: function(element) {
    return $(element).data('validator-name') || element.name;
}
```
and we can refactor finding by name as such:
```javascript
findByName: function( name ) {
    var escaped = this.escapeCssMeta( name );
    return $( this.currentForm ).find( "[data-validator-name=" + escaped + "], [name='" + escaped + "']" );
}
````

staabm:
@chrishaines changing one attribute for another doesnt make much difference?

I understood the ticket as getting rid of the name attribute completly, if possible.

chrishaines:
@staabm it maintains backwards compatibility and allows you to use the ```rules``` option for ```validate()```. It would also require minimum refactoring.

Overbryd:
[...]

PR #1648, opened 2015-11-13
Core: Don't require name attribute for validation: allow data-validator-name override

lithiumFlower:
The impetus for this is working with sensitive input fields, particularly creditcard data. This is data we never want sent to our servers under any circumstances. The easiest and cleanest way to do that is to not put a name attribute on the field.

Overriding the name attribute with a data-validator-attribute (and favoring it over name when it exists) allows leaving off the name attribute, maintains backwards compatibility, and is opt-in.

closes https://github.com/jzaefferer/jquery-validation/issues/659

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -56,6 +56,7 @@ $.extend( $.fn, {
 							// Insert a hidden input as a replacement for the missing submit button
 							hidden = $( "<input type='hidden'/>" )
 								.attr( "name", validator.submitButton.name )
+								.data( "validator-name", $.data( validator.submitButton, "validator-name" ) )
 								.val( $( validator.submitButton ).val() )
 								.appendTo( validator.currentForm );
 						}
@@ -128,14 +129,14 @@ $.extend( $.fn, {
 
 				// Remove messages from rules, but allow them to be set separately
 				delete existingRules.messages;
-				staticRules[ element.name ] = existingRules;
+				staticRules[ $.validator.getName( element ) ] = existingRules;
 				if ( argument.messages ) {
-					settings.messages[ element.name ] = $.extend( settings.messages[ element.name ], argument.messages );
+					settings.messages[ $.validator.getName( element ) ] = $.extend( settings.messages[ $.validator.getName( element ) ], argument.messages );
 				}
 				break;
 			case "remove":
 				if ( !argument ) {
-					delete staticRules[ element.name ];
+					delete staticRules[ $.validator.getName( element ) ];
 					return existingRules;
 				}
 				filtered = {};
@@ -256,7 +257,7 @@ $.extend( $.validator, {
 			}
 		},
 		onfocusout: function( element ) {
-			if ( !this.checkable( element ) && ( element.name in this.submitted || !this.optional( element ) ) ) {
+			if ( !this.checkable( element ) && ( $.validator.getName( element ) in this.submitted || !this.optional( element ) ) ) {
 				this.element( element );
 			}
 		},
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -937,8 +937,13 @@ $.extend( $.validator, {
 			return ( /radio|checkbox/i ).test( element.type );
 		},
 
+		getName: function( element ) {
```
lithiumFlower: Put this in a better place

src/core.js
```diff
@@ -128,14 +128,14 @@ $.extend( $.fn, {
 
 				// Remove messages from rules, but allow them to be set separately
 				delete existingRules.messages;
-				staticRules[ element.name ] = existingRules;
+				staticRules[ validator.getName( element ) ] = existingRules;
```
Arkni: `validator` is not defined

src/core.js
```diff
@@ -128,14 +128,14 @@ $.extend( $.fn, {
...
 				if ( argument.messages ) {
-					settings.messages[ element.name ] = $.extend( settings.messages[ element.name ], argument.messages );
+					settings.messages[ validator.getName( element ) ] = $.extend( settings.messages[ validator.getName( element ) ], argument.messages );
```
Arkni: The same here

src/core.js
```diff
@@ -128,14 +128,14 @@ $.extend( $.fn, {
...
 				}
 				break;
 			case "remove":
 				if ( !argument ) {
-					delete staticRules[ element.name ];
+					delete staticRules[ validator.getName( element ) ];
```
Arkni: And here as well

src/core.js
```diff
@@ -465,7 +465,7 @@ $.extend( $.validator, {
 
 				// Remove items from success list
 				this.successList = $.grep( this.successList, function( element ) {
-					return !( element.name in errors );
+					return !( this.getName( element ) in errors );
```
Arkni: `this` inside `$.grep` is the global window object

src/core.js
```diff
@@ -561,7 +561,7 @@ $.extend( $.validator, {
 		findLastActive: function() {
 			var lastActive = this.lastActive;
 			return lastActive && $.grep( this.errorList, function( n ) {
-				return n.element.name === lastActive.name;
+				return this.getName( n.element ) === this.getName( lastActive );
```
Arkni: `this` inside `$.grep` is the global window object

src/core.js
```diff
@@ -575,7 +575,7 @@ $.extend( $.validator, {
 			.not( ":submit, :reset, :image, :disabled" )
 			.not( this.settings.ignore )
 			.filter( function() {
-				var name = this.name || $( this ).attr( "name" ); // For contenteditable
+				var name = this.getName( this ) || $( this ).attr( "name" ); // For contenteditable
```
Arkni: `this` here is an `HTMLElement` instance represent the selected input

src/core.js
```diff
@@ -1356,11 +1361,11 @@ $.extend( $.validator, {
 			var previous = this.previousValue( element ),
 				validator, data, optionDataString;
 
-			if ( !this.settings.messages[ element.name ] ) {
-				this.settings.messages[ element.name ] = {};
+			if ( !this.settings.messages[ validator.getName( element ) ] ) {
```
Arkni: `validator` at this point is `undefined` and only initialized in line 1377

lithiumFlower:
@Arkni Thanks for reviewing.  I read the contribution section (my apologies for not following at first), and setup grunt etc with the tests.  I've rebased and combined/renamed the commits to reflect the guidelines, as well as fixed the errors in the original.

src/core.js
```diff
@@ -56,6 +56,7 @@ $.extend( $.fn, {
 							// Insert a hidden input as a replacement for the missing submit button
 							hidden = $( "<input type='hidden'/>" )
 								.attr( "name", validator.submitButton.name )
+								.data( "validator-name", $.validator.getName( validator.submitButton ) )
```
lithiumFlower: use `$.data`

Arkni:
> @Arkni Thanks for reviewing

You're welcome :)

The PR LGTM, but I think the maintainer(s) will not land it without a unit test.

staabm:
the feature request in #659 aims to drop the name attribute, without introducing alternative.
I am still not convinced that introducing another attribute which is used as a name makes sense.

Arkni:
> the feature request in #659 aims to drop the name attribute, without introducing a alternative.

This will work if you're using the declarative way (inline attributes).

There is another option is, to rely on the `id` attribute, but not every dev adds the `id` to the element.
Another idea is by using a selector, then after getting the element we can add an attribute to it, something like `data-validate-name = "username"` contains the same name used when initializing the plugin _(with JS)_, so we can do something like this:
```js
$( "form" ).validate( {
	rules: {
		username: {
			// Using a selector to identify the element
			selector: "some selector", // can be a class name, an attribute ...
			required: true,
			minlength: 2
		}
	}
} );
```
but this is not a favorable solution as you said `without introducing an alternative`.

But I think removing the requirement of `name` attribute is a BC break.

staabm:
> But I think removing the requirement of name attribute is a BC break.

right, this would require a new major version.

> The impetus for this is working with sensitive input fields, particularly creditcard data. This is data we never want sent to our servers under any circumstances. The easiest and cleanest way to do that is to not put a name attribute on the field.

@lithiumFlower maybe you shoud describe this a bit more. how do you get your creditcard data to the server, in case you have a field without a name?
Lots of browser only allow correct autocomplete for CC data, when it has a proper name.. do we have other examples except CC which make sense?

lithiumFlower:
@staabm Thanks for addressing the PR! I'll lay out my case for it and would love to hear you input on it.

We don't store or send credit card data to our servers - it's managed securely by Braintree, and we use a secure token to process payments with Braintree.  It's standard for Braintree (as well as Stripe) that the name attribute shouldn't be included on credit card fields.

From stackoverflow questions, blog posts, and processor docs, it looks like this is an issue a lot of people have to get around. In fact, Stripe used to use jquery-validate and dynamically added and removed the name attribute at the last minute (this is one of the common hack solutions). They've since written their own validation library specifically for sensitive credit card fields (which uses a data-stripe attribute):
http://formvalidation.io/examples/validating-custom-stripe-form/

The PR solution maintains backwards compatibility, is ready to go, and addresses these name issues: 
1. a unique field is needed
2. the name field is unavailable for use for whatever reason (ours being sensitive data being sent to server)
3. a field that isn't coupled to any constraints, whether from other libraries or html specification use

Do you think there is a better, more practical solution to address these points that could be implemented anytime soon?

staabm:
An alternative approach is one of
- used markup based rules
- use api based rules, see http://jqueryvalidation.org/rules

staabm:
@lithiumFlower does one of the mentioned alternatives above work for you?

Will close the Pr for now. We can re-open in case we get to a point where current api is not sufficient.

PR #1651, opened 2015-11-16
Core: Extract the file name without including "C:\fakepath\"

### LINKED ISSUE #1615: C:\fakepath\ is added to length when checking the length of a file type
C:\fakepath\ is added in to the value field of file inputs. This should be stripped out when validated. The validators affected are maxlength, minlength, rangelength, range, min, and max. My current solution is to add `b=a(c).attr('type')=='file'?b.replace(/C:\fakepath\/i,'');` to each of the validators. However, I know this is not an elegant solution and was wondering if one the maintainers could point me in the right direction.

Testing Notes:
test form: https://jsfiddle.net/zp423Lup/6/
test file to downloaded file: https://www.owasp.org/images/5/52/OWASP_Testing_Guide_v4.pdf
note that the length of the file name is 26 characters and the maxlength is 30 characters so it should pass.

Browsers affected: latest version of Chrome and Safari
Plug-in Versions affected: v1.13.0 - latest

Arkni:
For historical reasons, the value IDL attribute prefixes the file name
with the string "C:\fakepath\". As a result of this, this fix will extract
the file name from the value IDL attribute in a backwards-compatible way.

For more details, see:
 - http://www.w3.org/TR/html5/forms.html#dom-input-value-filename
 - http://www.w3.org/TR/html5/forms.html#fakepath-srsly

Fixes #1615

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -761,6 +761,55 @@ test( "elementValue() finds radios/checkboxes only within the current form", fun
 	ok( !v.elementValue( foreignRadio ) );
 } );
 
+test( "elementValue() returns the file input's name without the prefix 'C:\\fakepath\\' ", function() {
+	var v = $( "#userForm" ).validate(),
+
+		// A fake file input
+		fileInput = {
+			name: "fakeFile",
+			type: "file",
+			files: {},
+			nodeName: "INPUT",
+			value: "C:\\fakepath\\somefile.txt",
+			form: $( "#userForm" )[ 0 ],
+			hasAttribute: function() { return false; },
+			getAttribute: function( name ) {
+				if ( name === "type" ) {
+					return "file";
+				}
+
+				return undefined;
+			},
+			setAttribute: function() {}
+		};
+
+	v.defaultShowErrors = function() {};
+	v.validationTargetFor = function() {
+		return fileInput;
+	};
+
+	equal( v.elementValue( fileInput ), "somefile.txt" );
+
+	$( fileInput ).rules( "add", {
+		minlength: 10
+	} );
+
+	ok( v.element( fileInput ), "The fake file input is valid (length = 12, minlength = 10)" );
+
+	fileInput.value = "C:\\fakepath\\file.txt";
+	ok( !v.element( fileInput ), "The fake file input is invalid (length = 8, minlength = 10)" );
+
+	$( fileInput ).rules( "remove" );
+	$( fileInput ).rules( "add", {
+		maxlength: 10
+	} );
+
+	ok( v.element( fileInput ), "The fake file input is valid (length = 8, maxlength = 10)" );
+
+	fileInput.value = "C:\\fakepath\\fakefile.txt";
+	ok( !v.element( fileInput ), "The fake file input is invalid (length = 12, maxlength = 10)" );
+} );
+
 test( "validating multiple checkboxes with 'required'", function() {
 	expect( 3 );
 	var checkboxes = $( "#form input[name=check3]" ).prop( "checked", false ),
--- a/src/core.js
+++ b/src/core.js
@@ -624,9 +624,9 @@ $.extend( $.validator, {
 		},
 
 		elementValue: function( element ) {
-			var val,
-				$element = $( element ),
-				type = element.type;
+			var $element = $( element ),
+				type = element.type,
+				val, idx;
 
 			if ( type === "radio" || type === "checkbox" ) {
 				return this.findByName( element.name ).filter( ":checked" ).val();
@@ -639,6 +639,31 @@ $.extend( $.validator, {
 			} else {
 				val = $element.val();
 			}
+
+			if ( type === "file" ) {
+
+				// Modern browser (chrome & safari)
+				if ( val.substr( 0, 12 ) === "C:\\fakepath\\" ) {
+					return val.substr( 12 );
+				}
+
+				// Legacy browsers
+				// Unix-based path
+				idx = val.lastIndexOf( "/" );
+				if ( idx >= 0 ) {
+					return val.substr( idx + 1 );
+				}
+
+				// Windows-based path
+				idx = val.lastIndexOf( "\\" );
+				if ( idx >= 0 ) {
+					return val.substr( idx + 1 );
+				}
+
+				// Just the file name
+				return val;
+			}
+
 			if ( typeof val === "string" ) {
 				return val.replace( /\r/g, "" );
 			}


Arkni:
Just remembered this and wondering why this should be added to the plugin if we have a `transformer` option that we can use for that.

_JUST WONDERING_
* What someone can do with the name of a file ?
* Why they want to validate its length ?

Finally, I think this PR should remain open until someone else claim about that.

@staabm what do you think ?

staabm:
With this change we only cut the `C:\fakepath\` prefix but leave the file-path and file-name as sent via the browser, right?

A user might validate the input file-name/path because of certain length restrictions in the db (e.g. `VARCHAR(255)`). He might also check the filename for certain specialchars, which the backend does not support (e.g. utf8 chars in filenames).

Arkni:
> With this change we only cut the C:\fakepath\ prefix but leave the file-path and file-name as sent via the browser, right?

The browser will send the file name prefixed by `C:\fakepath\`, so yes it only cuts that part and return the file name (as shown in unit test).

> A user might validate the input file-name/path because of certain length restrictions in the db (e.g. VARCHAR(255)).
> He might also check the filename for certain specialchars, which the backend does not support (e.g. utf8 chars in filenames).

Make sense. Thanks for the explanation :)

I've handled that in one of my project by generating a new name for the file (so the original name doesn't matter anymore, and sometimes using the title as name) and rely on the title that the user should enter.

> Also as described in #1615 the maxlength and minlength options do not work correctly because of the magic-prefix.

Yes, so the PR still valid :D

staabm:
@Arkni what do you think about adding a test for min or max length check for a file input?

Arkni:
> @Arkni what do you think about adding a test for min or max length check for a file input?

Make sense as the issue we are trying to solve was about that.
I will try to work on it and see what can be done.

Arkni:
@staabm please double check.

If you have a better way to do the test, please let me know.

test/test.js
```diff
@@ -761,6 +761,52 @@ test( "elementValue() finds radios/checkboxes only within the current form", fun
...
+			maxlength: 10
+		};
+	};
+
+	ok( v.check( fileInput ), "The fake file input is valid (length = 8, maxlength = 10)" );
+
+	fileInput.value = "C:\\fakepath\\fakefile.txt";
+	ok( !v.check( fileInput ), "The fake file input is invalid (length = 12, maxlength = 10)" );
+
+	$.fn.rules = oldRules;
```
staabm: why resetting to `oldRules` and `oldvalidationTargetFor`?

test/test.js
```diff
@@ -761,6 +761,52 @@ test( "elementValue() finds radios/checkboxes only within the current form", fun
...
+		return {
+			minlength: 10
+		};
+	};
+
+	v.validationTargetFor = function() {
+		return fileInput;
+	};
+
+	ok( v.check( fileInput ), "The fake file input is valid (length = 12, minlength = 10)" );
```
staabm: this is not documented api, you should use `valid` instead

test/test.js
```diff
@@ -761,6 +761,52 @@ test( "elementValue() finds radios/checkboxes only within the current form", fun
...
+			name: "fakeFile",
+			type: "file",
+			files: {},
+			nodeName: "INPUT",
+			value: "C:\\fakepath\\somefile.txt",
+			form: $( "#userForm" )[ 0 ],
+			hasAttribute: function() { return false; }
+		},
+		oldRules = $.fn.rules,
+		oldvalidationTargetFor = v.validationTargetFor;
```
staabm: `oldValidationTargetFor`

test/test.js
```diff
@@ -761,6 +761,52 @@ test( "elementValue() finds radios/checkboxes only within the current form", fun
...
+
+	equal( v.elementValue( fileInput ), "somefile.txt" );
+
+	$.fn.rules = function() {
```
staabm: why do you monkey patch the rules into the object instead of using regular public api?

test/test.js
Arkni: My bad, sorry I'll change it.

test/test.js
Arkni: I was getting strange behavior using the public api directly because I'm using a fake input not a real one, but right now, I get an idea to solve that. will push again.

test/test.js
Arkni: in order to not affect other tests, I guess :)

test/test.js
Arkni: Is there a way to validate an element without showing the error ?
I guess the only the solution I have is to use undocumented api.

staabm: why does it matter if it is shown or not?

Arkni: Because I get this error, and it's caused because of `defaultShowErrors()`:
```
at http://127.0.0.1:34545/test/test.js:764:1: Cannot read property 'createDocumentFragment' of undefined@ 8 ms
Source: 	
TypeError: Cannot read property 'createDocumentFragment' of undefined
    at createSafeFragment (http://127.0.0.1:34545/lib/jquery-1.11.1.js:5274:22)
    at Function.jQuery.extend.buildFragment (http://127.0.0.1:34545/lib/jquery-1.11.1.js:5545:11)
    at jQuery.fn.extend.domManip (http://127.0.0.1:34545/lib/jquery-1.11.1.js:5910:22)
    at jQuery.fn.extend.after (http://127.0.0.1:34545/lib/jquery-1.11.1.js:5755:15)
    at jQuery.each.jQuery.fn.(anonymous function) [as insertAfter] (http://127.0.0.1:34545/lib/jquery-1.11.1.js:5987:35)
    at Object.$.extend.showLabel (http://127.0.0.1:34545/dist/jquery.validate.js:893:12)
    at Object.$.extend.defaultShowErrors (http://127.0.0.1:34545/dist/jquery.validate.js:829:10)
    at Object.$.extend.showErrors (http://127.0.0.1:34545/dist/jquery.validate.js:492:10)
    at Object.$.extend.element (http://127.0.0.1:34545/dist/jquery.validate.js:466:9)
    at Object.<anonymous> (http://127.0.0.1:34545/test/test.js:798:9)
```

Arkni: And the input I'm using is not a real one, just a normal Object.

test/test.js
Arkni: Fixed.

test/test.js
staabm: @Arkni each test should live in its sandbox and the testframework is responsible for this kind of cleanup (I am not sure though qunit supports my assumption) ;)

staabm: I dont remember other unit tests doing such cleanup, so I am pretty sure we would have a lot of problems in case qunit would not already clean those things up (or boots a fresh environment for each test)

Arkni: > I dont remember other unit tests doing such cleanup, so I am pretty sure we would have a lot of problems in case qunit would not already clean those things up (or boots a fresh environment for each test)

I don't see such thing in their documentation (If I didn't miss something).
I guess we should ask someone from the core team for an answer, I don't know if jzaefferer has the time to do so.

Arkni: But we should always play it safe ;)

Arkni: OK, get ride or monkey patch `rules()` and used public api for validation :)

PR #1657, opened 2015-12-01
Core: Adding a way to pass method name to remote

wojwal:
This allows reusing remote as custom method via addMethod. Without this fix it will show standard "remote" message - instead of custom message from addMethod. 

Sample usage:
```javascript
	$.validator.addMethod('workemail', function(value, element, param) {
		return $.validator.methods.remote.call(this, value, element, {
			url: "validate/workemail/"
		}, 'workemail');
	}, 'work email custom message');
```

And in HTML I'm just using data-rule-workemail='true' attribute for input

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1029,11 +1029,11 @@ $.extend( $.validator, {
 			}
 		},
 
-		previousValue: function( element ) {
+		previousValue: function( element, method ) {
 			return $.data( element, "previousValue" ) || $.data( element, "previousValue", {
 				old: null,
 				valid: true,
-				message: this.defaultMessage( element, { method: "remote" } )
+				message: this.defaultMessage( element, { method: method } )
 			} );
 		},
 
@@ -1379,19 +1379,21 @@ $.extend( $.validator, {
 		},
 
 		// http://jqueryvalidation.org/remote-method/
-		remote: function( value, element, param ) {
+		remote: function( value, element, param, method ) {
 			if ( this.optional( element ) ) {
 				return "dependency-mismatch";
 			}
 
-			var previous = this.previousValue( element ),
+			method = typeof method === "string" && method || "remote";
+
+			var previous = this.previousValue( element, method ),
 				validator, data, optionDataString;
 
 			if ( !this.settings.messages[ element.name ] ) {
 				this.settings.messages[ element.name ] = {};
 			}
-			previous.originalMessage = previous.originalMessage || this.settings.messages[ element.name ].remote;
-			this.settings.messages[ element.name ].remote = previous.message;
+			previous.originalMessage = previous.originalMessage || this.settings.messages[ element.name ][ method ];
+			this.settings.messages[ element.name ][ method ] = previous.message;
 
 			param = typeof param === "string" && { url: param } || param;
 			optionDataString = $.param( $.extend( { data: value }, param.data ) );
@@ -1414,7 +1416,7 @@ $.extend( $.validator, {
 					var valid = response === true || response === "true",
 						errors, message, submitted;
 
-					validator.settings.messages[ element.name ].remote = previous.originalMessage;
+					validator.settings.messages[ element.name ][ method ] = previous.originalMessage;
 					if ( valid ) {
 						submitted = validator.formSubmitted;
 						validator.prepareElement( element );
@@ -1424,7 +1426,7 @@ $.extend( $.validator, {
 						validator.showErrors();
 					} else {
 						errors = {};
-						message = response || validator.defaultMessage( element, { method: "remote", parameters: value } );
+						message = response || validator.defaultMessage( element, { method: method, parameters: value } );
 						errors[ element.name ] = previous.message = message;
 						validator.invalid[ element.name ] = true;
 						validator.showErrors( errors );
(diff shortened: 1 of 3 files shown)


wojwal:
Hi, just wanted to get an update on this pull-request? :)

staabm:
@wojwal thx for the PR. It somehow slipped thru :-)

I like it. Could you rebase it and add unit tests?

Arkni:
I'm not against this approach, but can't we achieve the same result if we use: `data-rule-remote="validate/workemail/"` and `data-msg-remote="work email custom message"` ?

Please correct me if I misunderstood something.

wojwal:
@staabm - I just rebased the code - so that part is ready. As for test units - I reviewed "test" folder, but it will take me some time to understand, how to do it properly - I'll do it when I got some spare time. 

@Arkani - Not sure I guess your solution will work, but there is much more you can do with solution I proposed. For example:
 - you can have multiple custom methods in one page that uses remote - I think in the way you suggest you can have only 1 remote method with 1 defined message?
 - you can pass custom parameters as "data" to XHR requet 
 - I use Javascript localization - so error message I passed to validator is additionally translated via Javascript - you wouldn't be able to do it with data-msg-remote attribute
 - HTML-wise - it's simpler just to have single attribute: "data-rule-workemail='params'" ;-)

Arkni:
So your idea is using the `remote` rule inside other custom rules and use the error message of your rule as the one for the `remote` method. I like it :)

Feel free to tag me whenever you need any help regarding the unit test :)

>By the way my username is @Arkni :D

src/core.js
```diff
@@ -1390,19 +1390,21 @@ $.extend( $.validator, {
...
 
 		// http://jqueryvalidation.org/remote-method/
-		remote: function( value, element, param ) {
+		remote: function( value, element, param, method ) {
 			if ( this.optional( element ) ) {
 				return "dependency-mismatch";
 			}
 
-			var previous = this.previousValue( element ),
+			method = typeof method === "string" ? method : "remote";
```
Arkni: Maybe this can be changed to:
```js
method = method || "remote";
```

wojwal:
Hi guys ;)

I've added tests for this, which includes:
- checking error message passed to addMethod
- additionally passing & requiring special parameter to XHR via data-rule-workemail attribute

@Arkni - as for your suggestion - I'm fine with either case. Your suggestion is just little less strict (not checking if value is string) - but thats not a big deal...

Arkni:
> as for your suggestion - I'm fine with either case. Your suggestion is just little less strict (not checking if value is string) - but thats not a big deal...

I know, but what will happen if a user passed an empty String (in a test for example) ?
So for a strict check, you could use:
```js
method = typeof method === "string" && method || "remote";
```

Otherwise, LGTM :)

wojwal:
I just commited that change ;)

wojwal:
HI guys, pls merge my changes... Anything missing? 

Thanks

staabm:
@wojwal please rebase the branch against master. After I merge some other PRs today it is no longer mergeable. Will merge after your rebase ASAP. thanks.

wojwal:
@staab rebased ;)

staabm:
@wojwal thx

Maks3w:
Just note this introduce a BC Break for those developers with custom validation functions.

This should be highlighted in the CHANGELOG

Arkni:
> Just note this introduce a BC Break ...

How so ?

Maks3w:
If the additional argument is not provided then `method` is `undefined`

When `method: undefined` is sent to `defaultMessage` app blow up because cannot do `method.charAt`

Maks3w:
So if previousValue `method` is not set then the value should fallback to `"remote"`

staabm:
@Maks3w please open a new issue

Arkni:
> If the additional argument is not provided then method is undefined

No, that's not true. If the argument is not provided, then `method` will be set to `remote` prior to any call to `previousValue`, see https://github.com/jzaefferer/jquery-validation/blob/73c173198c2ab225ecb0713c55f509b30acbb9c0/src/core.js#L1426-L1428

Maks3w:
> Just note this introduce a BC Break for those developers with **custom validation functions**.

Arkni:
Sorry, but I still didn't get it.

Maybe you can open an issue and provide a simplified example (in jsfiddle or something else).
Thanks!

Maks3w:
No need of jsfiddle `previousValue` is public and their signature must not change in unsafe way.

`publicFunction(oldArg, newArg) { if (newArg === undefined) newArg = defaultValue }`

Arkni:
Ah, I see. Thanks for the explanation.

This should be fixed as soon as possible, minor releases should never introduce breaking changes.

staabm:
@Arkni I would like to release 15.1 next week, could you provide a fix for this until then?

Maks3w:
One note defaultmessage has changed his behavior and the second argument must be an object. I'll put the note on the original PR if I found,

Maks3w:
https://github.com/jzaefferer/jquery-validation/pull/1644/files#r66706223

Arkni:
> @Arkni I would like to release 15.1 next week, could you provide a fix for this until then?

Sure!

PR #1658, opened 2015-12-02
BIC: allow digits 1-9 in second place of location

stefanb:
It seems that the BIC specification (https://en.wikipedia.org/wiki/ISO_9362 ) was interpreted too strictly, only allowing digits 1 and 2 while those 2 just have a special meaning, but all 1-9 are allowed.

Example: SSKNDE77XXX - http://bank-code.net/swift-code/SSKNDE77XXX.html (Sparkasse Nuernberg)

### DIFF
--- a/test/methods.js
+++ b/test/methods.js
@@ -760,7 +760,6 @@ test( "bic", function() {
 	ok( !method( "PBNKDEFFXXX1" ), "Invalid BIC: too long" );
 	ok( !method( "1BNKDEFF" ), "Invalid BIC: invalid digit" );
 	ok( !method( "PBNKDE1F" ), "Invalid BIC: invalid digit" );
-	ok( !method( "PBNKDEF0" ), "Invalid BIC: invalid digit" );
 	ok( !method( "PBNKDEFO" ), "Invalid BIC: invalid char" );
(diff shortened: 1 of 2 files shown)


test/methods.js
```diff
@@ -760,7 +760,7 @@ test( "bic", function() {
 	ok( !method( "PBNKDEFFXXX1" ), "Invalid BIC: too long" );
 	ok( !method( "1BNKDEFF" ), "Invalid BIC: invalid digit" );
 	ok( !method( "PBNKDE1F" ), "Invalid BIC: invalid digit" );
-	ok( !method( "PBNKDEF3" ), "Invalid BIC: invalid digit" );
+	ok( !method( "PBNKDEF0" ), "Invalid BIC: invalid digit" );
```
staabm: This assertion does no longer make sense.. (Duplicates with the next line)?

stefanb: Next line tests for letter "O", but testing for digit zero "0" is done few lines lower, so i will remove this, now redundant line.

stefanb:
Tests have been adjusted according to @staabm's comments.

staabm:
rebased and merged. thx @stefanb

PR #1659, opened 2015-12-03
Normalize min and max method parameters

rodrigoaguilera:


Arkni:
The min and max are not normalized at the first place, as they are used as well in comparing dates (min and max date).

Also, there are 7 failing tests.

rodrigoaguilera:
That's what I noticed in the failing tests. I guess is up to the developer to pass the right config and this PR can be closed.
Still feels weird to use that rule for dates and numbers though

Arkni:
> Still feels weird to use that rule for dates and numbers though

The date input has min and max attribute (http://www.w3.org/TR/html-markup/input.date.html) but not every browser has support for it (e.g. IE, Firefox and Safari. See http://caniuse.com/#feat=input-datetime).

Also it's already mentioned in the source code, see [src/core.js#L1092-L1101](https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L1092-L1101).

rodrigoaguilera:
Thank you :)

Arkni:
You're welcome :)

staabm:
@rodrigoaguilera can we also close https://github.com/jzaefferer/jquery-validation/issues/1660 then?

rodrigoaguilera:
Oh yeah, I forgot

PR #1662, opened 2015-12-09
Core: closes #1661

### LINKED ISSUE #1661: filled selector incorrect for select[multiple] elements
It appears that jquery-validation is giving incorrect results with the `:filled` selector for `<select multiple="multiple" />` elements.

I have created a [jsfiddle](http://jsfiddle.net/98xb8j5y/) to show an example.

It appears that the reason for this inconsistency is that the `:filled` selector is checking for [`return !!$.trim( "" + $( a ).val() );`](https://github.com/jzaefferer/jquery-validation/blob/45f3ac25c596764ba770aceb169406e19acfbe93/src/core.js#L193). 

`$( a ).val()` will return `null` on `select[multiple]` elements when they do not have a value. 

Thus: `!!$.trim( "" + null) == true`

In previous versions we were checking for `a.value` which would return `""` for empty `select[multiple]` elements. This would give us correct results here in this instance. Otherwise we might check if the value is null first to give accurate results.

kjprice:


staabm:
manually merged, thx.

kjprice:
Great, thanks @staabm. Any idea when this might make it to production?

staabm:
@kjprice I guess when time allows we can aim for a christmas gift

kjprice:
Haha. Sounds good. Thanks again @staabm.

PR #1665, opened 2015-12-15
a little advice

yalishizhude:
This is a great and convenient component.
I use it find one problem: it can't validate form elements which have the same "name" attribute.So I advise to modify the function "elements".
Make it possible to  validate elements with the same name.
3q~


Arkni:
Duplicate of #1330 
So maybe you should close this in favor of it.

PR #1672, opened 2015-12-17
add validation method rib_fr_simple

lingtalfi:
RIB is the national bank account key used in France.


staabm:
Thx for your contribution. Since this is a very specific method, I would suggest publishing it in a separate repo.

Thanks

PR #1673, opened 2015-12-18
Persian localization fix

EsiKhoob:
Hello, and thank you for this good plugin
I find a bug in Farsi (Persian) localization exist in messages_fa.js
The "min"&"max" methods messages in that file have bug. The Farsi "min"
`s message has this meaning: “please enter value greater that {0}
words.”
Phrase meaning "word" should be removed from Farsi message.


staabm:
merged, thx.

Issue #1663, opened 2015-12-12
Validation on input type number not working on Google Chrome


sachingadagi:
<h2><b> This breaks in Google chrome only </b></h2>
when an **input** element with _type=number_ is validated against 
``` required: true, digits:true``` and non-numeric values are supplied,  the plugin breaks.
Fiddle
[https://jsfiddle.net/5nx2qgu6/1/]

Thanks & Regards,
SG

staabm:
Not 100% sure, but it seems we only have a manual test for this case?
https://github.com/jzaefferer/jquery-validation/blob/d09a5ee19944c756311e6021871ae45f51901fd6/test/number.html

Is your error reproducible using the number.html file?

Arkni:
Duplicate of #1514 and #1611

staabm:
Will be handled in https://github.com/jzaefferer/jquery-validation/issues/1514 then, thx.

Issue #1611, opened 2015-10-12
Version 1.14.0 breaks number validation.


simone91:
Try to enter a letter in this sample form and submit it.
https://jsfiddle.net/2x9bbnu9/1/
https://jsfiddle.net/Lgkbok58/1/

Version 1.13.1 works correctly.
Version 1.14.0 does not report the error.

Arkni:
Duplicate of #1514

As I can tell, this is caused by this commit: 2ad0ca82daf091fc692018b38e01e896fd304015

PR #1679, opened 2016-01-08
Core: Required rule take precedence over number & digits rules

### LINKED ISSUE #1514: Required fields error message takes precedence of number type validation
Related to #1318, it appears the issue of the required field error message taking precedence over the number type validation has gotten more prominent as it is now visible in many environments.

Minimum test case (jQuery 1.11.0): http://jsfiddle.net/L9ycncrg/
Minimum test case (jQuery 2.1.3): http://jsfiddle.net/uqdqbofy/

Environments effected:
jQuery 1.11.0
jQuery Validation 1.14.0

Windows 7: Internet Explorer 10
OSX: Chrome 43.0.2357.132, Firefox, Safari 8.0.7 (10600.7.12)
iOS8: Chrome, Safari

In most browsers I've found that a required number field will continue to display "This field is required." instead of "Please enter a valid number." if alpha characters are inserted instead of numeric characters. If I add numbers that exceed min/max rules then the required field error is replaced by the min/max error.

I found it to work correctly when running Internet Explorer 9 on Windows 7.

I did not see any behavioral differences between the versions of jQuery.

In the minimum test cases follow these steps to reproduce:
1. Click submit to trigger validation
2. Type "aaa" in the field, notice that the message does not ask you to enter a number, but rather complains that the field is required.

capture

Arkni:
Ref #1611
Ref #1663
Fixes #1514

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -644,7 +644,7 @@ $.extend( $.validator, {
 			if ( type === "radio" || type === "checkbox" ) {
 				return this.findByName( element.name ).filter( ":checked" ).val();
 			} else if ( type === "number" && typeof element.validity !== "undefined" ) {
-				return element.validity.badInput ? false : $element.val();
+				return element.validity.badInput ? "NaN" : $element.val();
 			}
 
 			if ( element.hasAttribute( "contenteditable" ) ) {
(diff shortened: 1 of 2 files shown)


staabm:
Will land on monday

staabm:
merged in 40dae365edf6ec44abcd46c10473f0e9d429a13a 

thanks.

PR #1680, opened 2016-01-10
[Remote rule] Core: Validation fails to trigger when next field is already filled out

Arkni:
This change fix the bug when used with remote rule, it's already fixed
whith other rules in commit 737630e7888485882c8c65b0f8ac5328950be0ac

Ref #1508

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -289,7 +289,7 @@ $.extend( $.validator, {
 
 			if ( event.which === 9 && this.elementValue( element ) === "" || $.inArray( event.keyCode, excludedKeys ) !== -1 ) {
 				return;
-			} else if ( element.name in this.submitted || this.isValidElement( element ) ) {
+			} else if ( element.name in this.submitted || element.name in this.invalid ) {
 				this.element( element );
 			}
 		},
@@ -487,6 +487,7 @@ $.extend( $.validator, {
 			if ( $.fn.resetForm ) {
 				$( this.currentForm ).resetForm();
 			}
+			this.invalid = {};
 			this.submitted = {};
 			this.prepareForm();
 			this.hideErrors();
@@ -537,18 +538,6 @@ $.extend( $.validator, {
 
 		valid: function() {
 			return this.size() === 0;
-		},
-
-		// Check if the given element is valid
-		// return
-		//          true  If the field is valid
-		//         false  If the field is invalid
-		//     undefined  If the field is not validated yet.
-		//
-		// Note that this method assumes that you have
-		// already called `validate()` on your form
-		isValidElement: function( element ) {
-			return this.invalid[ element.name ] === undefined ? undefined : !this.invalid[ element.name ];
 		},
 
 		size: function() {
@@ -1431,7 +1420,7 @@ $.extend( $.validator, {
 						validator.prepareElement( element );
 						validator.formSubmitted = submitted;
 						validator.successList.push( element );
-						delete validator.invalid[ element.name ];
+						validator.invalid[ element.name ] = false;
 						validator.showErrors();
 					} else {
 						errors = {};
--- a/test/test.js
+++ b/test/test.js
@@ -48,6 +48,19 @@ $.mockjax( {
 	}
 } );
 
+$.mockjax( {
+	url: "issue1508.php",
+	response: function( settings ) {
+		if ( /abc/i.test( settings.data.val2 ) ) {
+			this.responseText = "false";
+		} else {
+			this.responseText = "true";
+		}
+	},
+	responseStatus: 200,
+	responseTime: 1
+} );
+
 // Asserts that there is a visible error with the given text for the specified element
 QUnit.assert.hasError = function( element, text, message ) {
 	var errors = $( element ).closest( "form" ).validate().errorsFor( element[ 0 ] ),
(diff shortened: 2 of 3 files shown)


test/test.js
```diff
@@ -48,6 +48,19 @@ $.mockjax( {
 	}
 } );
 
+$.mockjax( {
```
staabm: looks like this mocking request alters global state. will it affect other tests?

test/test.js
```diff
@@ -1427,34 +1440,59 @@ test( "don't revalidate the field when pressing special characters", function()
...
+			element.trigger( event );
+			element.focus();
+
+			// Change the content
+			element.val( value );
+			element.trigger( "keyup" );
+
+			// 'element' should be invalid
+			if ( remote ) {
+				setTimeout( function() {
```
staabm: why do we need the timeout?

test/test.js
Arkni: noop, every `mockajax` call has a different url, so it will result in different object listener. (see the previous mockajax calls, just before this one)

test/test.js
Arkni: `setTimeout(function(){...})` simply queues the code inside it to run once the current call stack is finished executing. So the assert will run after the remote call has complete. You can see that some of the `remote` rule's tests are based on this approach.

See these stackoverflow threads:
* http://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful
* http://stackoverflow.com/questions/19626680/is-settimeout-a-good-solution-to-do-async-functions-with-javascript

There is another solution by using `asyncTest`.

test/test.js
staabm: in this case I would prefer we would use a url which cannot conflict with other tests by mistake, e.g. adding a request param into the url with the issue id or something (just as a marker)

test/test.js
staabm: please use `asyncTest`

Arkni: I will have a look at it when time allows :)

test/test.js
Arkni: Make sense, I will update the PR ASAP :)

staabm: maybe we should even move the `$.mockjax` call into the test which is using it..?

test/test.js
staabm: as always, dont hurry ;). thx so far.

test/test.js
Arkni: It's the same thing, as `mockjax` add the handler to the global scope. So it doesn't matter if it's inside the test or outside.

But using outside has its benefits, maybe someone can use it to in his test.

test/test.js
Arkni: > please use asyncTest

`asyncTest` is deprecated and will be removed in the next major version of `Qunit`, from [Docs](http://api.qunitjs.com/QUnit.asyncTest/):
>This method is deprecated and will be removed in QUnit 2.0.0. Use QUnit.test() in conjunction with assert.async() instead.

maybe we should update all the tests that use `asyncTest`.

staabm: Please use asyncTest for now. We will upadte the entire codebase in a separate PR. Thx for the heads up.

Arkni:
@staabm 
PR updated!

test/test.js
```diff
@@ -48,6 +48,19 @@ $.mockjax( {
 	}
 } );
 
+$.mockjax( {
+	url: "issue1508.php",
```
Arkni: As discussed above, I changed the url as well :)

test/test.js
Arkni: Updated the url to `issue1508.php`

staabm:
Very nice, thx. Did you make sure the test fails without the patch applied?

Arkni:
> Did you make sure the test fails without the patch applied?

Honestly, I tested it manually with a demo I created locally. After checking, I found that the test doesn't fail without the patch.

I will take another look, and see what's wrong.

Arkni:
@staabm 
This will remain open until I have some free time to dig into the issue, so please add the `NEEDS TEST` label as a reminder.
Thanks :)

src/core.js
```diff
@@ -539,18 +539,6 @@ $.extend( $.validator, {
...
-		//          true  If the field is valid
-		//         false  If the field is invalid
-		//     undefined  If the field is not validated yet.
-		//
-		// Note that this method assumes that you have
-		// already called `validate()` on your form
-		isValidElement: function( element ) {
-			return this.invalid[ element.name ] === undefined ? undefined : !this.invalid[ element.name ];
-		},
-
```
Arkni: We don't need this method any more

test/test.js
```diff
@@ -1473,31 +1486,102 @@ test( "don't revalidate the field when pressing special characters", function()
...
+
+			// Tab to the next field
+			element.blur();
+			nextE.trigger( event );
+			nextE.focus();
 
-	// Fill the first element
-	element.val( "abc" );
+			// Make sure all events will be called before the bellow code
+			setTimeout( function() {
```
Arkni: `setTimeout( ... )` is required here in order to make sure that all events (the first remote call and the rendering) are called before the second call of the remote method.

Arkni:
I split up the tests into 2, one for the `remote` rule and the second one for other rules.

I didn't use the `asyncTest` as discussed above, because it doesn't matter if I use it or if I use `test` instead, I will need to use `setTimeout` to make sure we have a proper unit test.

staabm:
merged, thx.

Issue #1683, opened 2016-01-12
How to check password strength ?


NiranjanaK:
Hi @jzaefferer 
I need some more additional requirement to check password strength like combination of uppercase,lowercase,digit and special character. Is there any function defined for that?

staabm:
Nope there isnt.

As mentioned earlier, I dont think this makes sense to be bundled with the lib. Password strength rules are very different, depending on the problem domain.

XenRevo:
I recommend [this](https://timseverien.com/passy/).

Joyrex:
Seems the library's author (at one point) disagreed with that stance: https://github.com/jzaefferer/jquery-validation.password

PR #1684, opened 2016-01-12
Additional: added password strength method.

NiranjanaK:
Hi @jzaefferer 
I implemented the password strength method, Please take a look. 

It's regarding the issue https://github.com/jzaefferer/jquery-validation/issues/1683


src/additional/passwordstrength.js
```diff
@@ -0,0 +1,3 @@
+$.validator.addMethod("pwcheck", function(value) {
```
staabm: should also be named `password-strength`

staabm:
please add a unit test

staabm:
on a second though: I guess nearly every application/corporation has different requirements/rules on password security.

I dont think this should be part of the library.

staabm:
@NiranjanaK please discuss the above mentioned issues at first. the PR will not be merged until we reached a consent.

NiranjanaK:
@staabm Thanku for your reviews.

staabm:
As mentioned earlier, I dont think this makes sense to be bundled with the lib. Password strenght rules are very different, depending on the problem domain.

PR #1687, opened 2016-01-15
Composer: Adds composer.json for package registration

augustpascual-mse:


staabm:
Composer is a package manager for php based packages which doesnt fit for this lib. Why do you want it in there?

augustpascual-mse:
Yes, but it's not only a package manager. Composer allows you to declare libraries in your project. It would be nice if we support people to download this lib when they already have composer installed. There are lots of packages in composer which are not PHP such as jQuery and have 600k+ download using composer.

staabm:
jquery and other js libs are not provided by the source repos but by so called shim reposes which distribute whatever think makes sense.

I dont think we should invest time to support another package manager, but concentrate on actual library features.

feel free to create such a shim repos on your own. thanks for the suggestion.

PR #1692, opened 2016-01-19
Core: Moves some element to settings to have users options when ignoring

### LINKED ISSUE #1691: Users are not allowed to not include some elements to ignore ie. disabled
I see from line 604-607 in jquery.validate.js that it's not including ignore settings. Also it's not including disabled, submit, reset, image as hard coded.
    return $( this.currentForm )
    .find( "input, select, textarea, [contenteditable]" )
    .not( this.settings.ignore )

If the user needs to include hidden fields, they can just put blank in the ignore settings like $("#signupForm").validate({ ignore : "" }); which is good. But I ran into some use cases that I'd like to validate disabled text fields. The only way to do that is to change the core which is a bad practice.

augustpascual-mse:
Fixes #1691

I see from line 604-607 in jquery.validate.js that it's not including ignore settings. Also it's not including disabled, submit, reset, image as hard coded.
    return $( this.currentForm )
    .find( "input, select, textarea, [contenteditable]" )
    .not( this.settings.ignore )

 If the user needs to include hidden fields, they can just put blank in the ignore settings like $("#signupForm").validate({ ignore : "" }); which is good. But I ran into some use cases that I'd like to validate disabled text fields. The only way to do that is to change the core which is a bad practice. 

 The changes proposed that we should treat :submit, :reset, :image, :disabled as settings.ignore. If the user needs to include disabled, they can just pass that as a parameter similar to hidden fields.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -248,7 +248,7 @@ $.extend( $.validator, {
 		errorContainer: $( [] ),
 		errorLabelContainer: $( [] ),
 		onsubmit: true,
-		ignore: ":hidden",
+		ignore: ":hidden, :submit, :reset, :image, :disabled",
 		ignoreTitle: false,
 		onfocusin: function( element ) {
 			this.lastActive = element;
@@ -585,7 +585,6 @@ $.extend( $.validator, {
 			// Select all valid inputs inside the form (no submit or reset buttons)
 			return $( this.currentForm )
 			.find( "input, select, textarea, [contenteditable]" )
-			.not( ":submit, :reset, :image, :disabled" )
 			.not( this.settings.ignore )
 			.filter( function() {
 				var name = this.name || $( this ).attr( "name" ); // For contenteditable
--- a/src/core.js
+++ b/src/core.js
@@ -248,7 +248,7 @@ $.extend( $.validator, {
 		errorContainer: $( [] ),
 		errorLabelContainer: $( [] ),
 		onsubmit: true,
-		ignore: ":hidden, :submit, :reset, :image, :disabled",
+		ignore: ":hidden, :disabled",
 		ignoreTitle: false,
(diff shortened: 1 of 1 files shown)


Arkni:
Instead of changing the core, have you considered using another solution ?
Maybe you can add a class to the disabled inputs you want to validate and then exclude them from the `ignore` setting. Something like that:
```js
ignore: ":disabled:not(.dont_ignore)"
```

Also, your solution is not proper. Instead of moving the 4 selectors (`:submit, :reset, :image, :disabled`) to `ignore` definition, you could only move the `:disabled` selector.

staabm:
this change looks like a massive BC break and therefore cannot be merged as is. feel free to discuss other solution to your problem though.

augustpascual-mse:
@Arkni 
1. I tried adding a class to the disabled inputs and it didn't work. Maybe I'm doing something wrong but I just test it really quick but even if it works, isn't it kinda hacky solution? I would say the proposed changes is better in terms of maintainability. 

2. Yeah, I was thinking of just moving :disabled selector. The reason I included everything is since we are ignoring those things, why not put it to settings? I think either way is fine but for now, let me update the pull request to only move :disabled. 

@staabm I don't think I understand it. What do you mean its massive and cannot be merged? The proposed changes is a simple solution and there is no conflict.

Arkni:
1. Sorry my fault, I forgot that the disabled inputs are filtered first and then filtered using the `ignore` setting. That's why it's not working.
2. Buttons and images are not meant to be validated, so it doesn't make sense to include them.

staabm:
> I don't think I understand it. What do you mean its massive and cannot be merged? The proposed changes is a simple solution and there is no conflict.

any project which used the `ignore` option before the PR needs to be adjusted when this gets merged.

PR #1700, opened 2016-02-01
Core: Fix TypeError in `showLabel` when using `groups` settings

Arkni:
The `this` inside `$.each()` refers to the current value of the
`Objcet|Array` we iterate over, not the plugin instance.

Ref https://github.com/jzaefferer/jquery-validation/pull/1614#discussion_r51156526

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -843,7 +843,7 @@ $.extend( $.validator, {
 		},
 
 		showLabel: function( element, message ) {
-			var place, group, errorID,
+			var place, group, errorID, v,
 				error = this.errorsFor( element ),
 				elementID = this.idOrName( element ),
 				describedBy = $( element ).attr( "aria-describedby" );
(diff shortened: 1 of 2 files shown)


test/messages.js
```diff
@@ -20,6 +20,7 @@ test( "group error messages", function() {
 
 	var form = $( "#dateRangeForm" );
 	form.validate( {
+		errorElement: "span",
```
Arkni: This way we can ensure that the line 909  will be called.

src/core.js
```diff
@@ -884,11 +884,10 @@ $.extend( $.validator, {
...
 					// If the error is a label, then associate using 'for'
 					error.attr( "for", elementID );
-				}
 
-				// If the element is not a child of an associated label, then it's necessary
-				// to explicitly apply aria-describedby
-				else if ( error.parents( "label[for='" + this.escapeCssMeta( elementID ) + "']" ).length === 0 ) {
+					// If the element is not a child of an associated label, then it's necessary
+					// to explicitly apply aria-describedby
+				} else if ( error.parents( "label[for='" + this.escapeCssMeta( elementID ) + "']" ).length === 0 ) {
```
Arkni: This fix a CS issue reported in https://github.com/jzaefferer/jquery-validation/issues/1618#issuecomment-159093885

staabm:
awesome thx. confirmed that the error gone away!

Arkni:
You're welcome :) Thanks for confirming :)

Issue #1707, opened 2016-02-10
[ equalTo ] bind `blur` event only once to the target of `equalTo` rule


Arkni:
From #1704:
* `@staabm`:

>we should use .one()[1] to fix that TODO. but this will also not clear the event listener on destroy
>
>[1] http://api.jquery.com/one/

* `me`:

>> we should use .one()[1] to fix that TODO.

>~~I think it's the same as using `.off( ... ).on( ... )` but more readable.~~
It's not equivalent to `off( ... ).on( ... )` in the use case of `equalTo` rule. The event should still there in order to validate the element (which `equalTo` rule was attached to) after every focus out of the target.
IMO, the goal of that TODO is to bind the event once for all without unbinding it every time.
~~Anyway, until there is a better solution, using `one( ... )` makes more sense here~~.

>> but this will also not clear the event listener on destroy

>We have to add it, then.

---
The problem here is, every time the element was validated, the event was (un)binding from|to `target`.
Even if we use `.one( ... )`, the only benefit is making the code more readable and simple.

//Cc @staabm

Arkni:
Doing some testing today I noticed that, when using `.one( ... )`, the number of handlers attached to `blur.validate-equalTo` event of the target incremented every time the element was validated (`onkeyup`). So using `one( ... )` is not an option here.
Following is a screen-cast of the result when using `one( ... )` to bind the blur event:
![anim](https://cloud.githubusercontent.com/assets/6888059/12980332/f400e198-d0d3-11e5-8a7b-15968e9fa3f5.gif)

staabm:
thx for testing. seems we have to leave it as is then.

Arkni:
There is a work around for fixing that, we can add a class `equalTo-blur` for example to `target`.
with this, we can avoid binding|unbinding the event every time:
```js
if ( this.settings.onfocusout )
    target.not( ".equalTo-blur" ).addClass( "equalTo-blur" ).on( "blur.validate-equalTo", function() {
        $( element ).valid();
    } );
}
```

or even better:
```js
if ( this.settings.onfocusout && target.not( ".equalTo-blur" ).length ) {
    target.addClass( "equalTo-blur" ).on( "blur.validate-equalTo", function() {
        $( element ).valid();
    } );
}
```
This will give us a way to unbind the `blur` event when destroying the plugin.

WDYT ?

staabm:
ohh this sounds like we could kill two birds with one stone. awesome.

[...]

Issue #1704, opened 2016-02-08
Feature request: Remove validator


RoBorg:
It would be useful to have an option to remove the validator from a form - I think all it needs to do it remove the data and unbind the events.

At the moment I'm doing this outside the plugin like this:

```
form.add(form.find(':submit, input, select, option, textarea')).off('.validate .validate-equalTo');
form.removeData('validator');
```

nebrot:
With 1.14 there is an "destroy"-method, see #1392

staabm:
netbrot had the right answer. will close this issue now.

feel free to ask if you have followup questions.

RoBorg:
Awesome, thanks!
A couple of small questions:

1. Can it get added to the documentation?
2. The equalTo method adds a "blur.validate-equalTo" event, which I don't believe will be unbound, is that right?

staabm:
1. you are more than welcome to provide the docs as a PR in https://github.com/jzaefferer/validation-content ;-)

2. I guess you are right. there is a open todo in the code[1].

[1] https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L1371

staabm:
we should use `.one` to fix that: http://api.jquery.com/one/

but this will also not clear the event listener on `destroy`

Arkni:
> we should use .one()[1] to fix that TODO.

I think it's the same as using `.off( ... ).on( ... )` but more readable.
IMO, the goal of that TODO is to bind the event once for all without unbinding it every time.
Anyway, until there is a better solution, using `one( ... )` makes more sense here.

> but this will also not clear the event listener on destroy

We have to add it, then.

Arkni:
@staabm I updated my previous comment!
I will create a new issue to discuss the first point.

staabm:
I guess we need to discuss how we can clear the listener on `destroy` - so how to find the elements on which this elements have been registered...

Arkni:
@staabm 
see https://github.com/jzaefferer/jquery-validation/issues/1707#issuecomment-182926873

PR #1709, opened 2016-02-11
Core: Bind the `blur` event just once in `equalTo` rule

Arkni:
bind the event just once, avoiding the unbind-rebind overhead.
Also, unbind it when destroying the plugin.

Ref #1704
Ref #1707

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1061,7 +1061,10 @@ $.extend( $.validator, {
 
 			$( this.currentForm )
 				.off( ".validate" )
-				.removeData( "validator" );
+				.removeData( "validator" )
+				.find( ".validate-equalTo-blur" )
+					.off( ".validate-equalTo" )
+					.removeClass( "validate-equalTo-blur" );
 		}
 
 	},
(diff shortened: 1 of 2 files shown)


staabm:
@Arkni do you consider adding a unit test? I am also ok to land this as is.

@RoBorg could you confirm that this patch fixes the issue you observed?

RoBorg:
@staabm yup that will fix it

Arkni:
@staabm I will try to extend `destroy`'s unit test.

Arkni:
I added some tests :)

staabm:
Nice. Did you make sure the tests fail without the patch?

Arkni:
Yep, I did :)

staabm:
merged thx. not sure why git wasnt able to fast forward it... required a merged commit..

Arkni:
I'm not sure why this happen.

The [Git's docs](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging#Basic-Merging) give us a good reason:
>Because the commit on the branch you’re on isn’t a direct ancestor of the branch you’re merging in, Git has to do some work. In this case, Git does a simple three-way merge, using the two snapshots pointed to by the branch tips and the common ancestor of the two.

Arkni:
If you want, you can reset the master and delete the last 2 commit (including the merge commit) and I will create another PR on base of master.

staabm:
No worries, its fine

Issue #1706, opened 2016-02-09
Calling .rules() on an empty jQuery object results in an error


whefter:
(Latest version)

Calling e.g.

`$('form.validate .somenonexistantclass').rules('add', 'whatever')`

results in

`Uncaught TypeError: Cannot read property 'form' of undefined`

The problems seems to be (from a quick glance) that the rules() method does not check for `this.length` but assigns `this[0]` straight away.

PR #1710, opened 2016-02-12
Tests: Calling .rules() on an empty jQuery object results in an error.

staabm:
Refs #1706

### DIFF
--- a/test/rules.js
+++ b/test/rules.js
@@ -396,3 +396,10 @@ test( "rules(), normalizer", function() {
 		return err.name === "TypeError" && err.message === "The normalizer should return a string value.";
 	}, "This should throw a TypeError exception." );
 } );
+
+test( "rules() - on empty jquery set #1706", function() {
+	var element = $( "#firstname .mynonexistantclass" );
+
+	var result = element.rules( "add", "whatever" );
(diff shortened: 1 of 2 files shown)


staabm:
@whefter could you confirm the fix?

test/rules.js
```diff
@@ -396,3 +396,10 @@ test( "rules(), normalizer", function() {
 		return err.name === "TypeError" && err.message === "The normalizer should return a string value.";
 	}, "This should throw a TypeError exception." );
 } );
+
+test( "rules() - on empty jquery set #1706", function() {
+	var element = $( "#firstname .mynonexistantclass" );
+
+	var result = element.rules( "add", "whatever" );
+	equal( result, null, "can work on an empty set without a js error" );
```
Arkni: I guess using `strictEqual` is more appropriate here:
```js
strictEqual( result, undefined, "can work on an empty set without a js error" );
```

Arkni:
A rebase is needed here.

staabm:
rebased and adjusted per feedback @Arkni.

just leveled my git knowledge up, yay :100:

Arkni:
:+1:

Issue #1669, opened 2015-12-17
Remote validation deletes previous validation errors


suclogger:
Similar to #1375 , #12  and #512 .

Here is my test code.
using jQuery 1.9.0 and jQuery validation 1.14.0
```
<!DOCTYPE html>
<html>
<head>
    <script src="jquery.js"></script>
    <script src="jquery.validate.js"></script>
    <script type="text/javascript">
        $(function () {
            $("#submitForm").validate({
                rules: {
                    "requiredElement": "required",
                    "remoteElement": {
                        required: true,
                        remote: {
                            url: "a method returns true",
                            type: "post",
                            data: {
                                test: "not necessary"
                            },
                            // set async to false
                            async: false
                        }
                    }
                }
            });
            $("#testBtn").click(function () {
                console.log($("#submitForm").validate().form());
            });
        });
    </script>
</head>
<body>
<form id="submitForm">
    <div>
        it's a required element!
    </div>
    <input name="requiredElement"/>
    <div>
        it's a remote element!
    </div>
    <input name="remoteElement"/>
    <div>
        it's a button!
    </div>
    <input type="button" id="testBtn" value="test"/>
</form>
</body>
</html>
```

Every first time the test button is clicked, the validate result is not correct: 
**Although the required element `requiredElement` is left empty, but the validate result still indicates valid.**

After some digging,in `core.js` line 1424:
```
validator.prepareElement( element );
```

`prepareElement` is called after remote method return a valid result, which cause the former errors being erased.  

Maybe you would indicate #12 #issuecomment-14179880 
>but he uses async: false, which isn't supported and never will be.

Bu if we don't use ` async: false`, the result of `$("#submitForm").validate().form()` is still not correct because the valid result would return valid before the asynchronous method returns.

**I didn't see why we need to call the `prepareElement` after remote method returns true, and I also didn't see why we shouldn't use `async: false`.**

And as mentioned issues are all closed, so I open this new issue to describe the problem. 
Hoping to get an answer on how to correctly use the remote check.

staabm:
using async=false is discouraged because it affects browser performance and can make it even stall.

staabm:
per reporter no longer reproducible, see https://github.com/jzaefferer/jquery-validation/pull/1712#issuecomment-185533704

Issue #1375, opened 2015-01-14
Remote validation deletes previous validation errors


florianengel:
If one element on a page is validated using the remote validation method, the validation errors of previous elements get lost.
Since I wasn't able to provide a working example on jsfiddle or jsbin, which actually does a remote call, I put the demo here: http://bit.ly/14Xt9Da

If you leave all 3 input elements empty, the required rule for the first and third element works correctly, since the remote validation doesn't get triggered. If you fill something in the second field, only the validation message for the third one shows up. (the true.php returns true for every input).
This behavior occurs also in the latest version 1.13.1. 

During the remote success handling the function prepareElement() gets called, which will then call this.reset();
```javascript
		reset: function() {
			this.successList = [];
			this.errorList = [];
			this.errorMap = {};
			this.toShow = $( [] );
			this.toHide = $( [] );
			this.currentElements = $( [] );
		},
```
Particularly the this.errorMap = {} seems to cause the problem here. Till the validator reaches the statement, it holds all the previous errors.

jzaefferer:
Thank you for the report. The remote method isn't the only one causing issues like this. Calling `valid()` on any element directly also does that. A proper solution would track each element individually, instead of dumping everything into a single object and resetting that together.

c-krani:
Recently we encountered the same issue. I don't unterstand why prepareElement() gets called in the success-Method of the remote validation. None of the other valdiation methods does. 
Anybody knows the reason?

jzaefferer:
I'm sorry for the lack of activity on this issue. Instead of leaving it open any longer, I decided to close old issues without trying to address them, to longer give the false impression that it will get addressed eventually, especially after several years with no activity. It doesn't mean I'm abandoning the project, just that I'm unable to work through 200+ open issues with the little time I can afford to spend on this project.

To the reporter (or anyone else interested in this issue): If you're still affected by the same issue, please consider opening a new issue, with a testpage that demonstrates the issue with a current version of the plugin. Even better, make an attempt to fix the issue yourself, and improve the project by sending a pull request. This may seem daunting at first, but you'll likely learn some useful skills that you can apply elsewhere as well. And you can help keep this project alive. [We've documented how to do these things, too](https://github.com/jzaefferer/jquery-validation/blob/master/CONTRIBUTING.md). A patch is worth a thousand issues!

Issue #12, opened 2010-11-10
Remote validation kills local validation


cvwillegen:
I have a couple of controls that are validated via a remote option. The 'success' function calls validator.prepareElement(element), which clears the error list.

I've tracked this bug down by changing a line in jquery.validate.js:918 (the start of the 'remote' function). When I return 'true' there, all the local errors show up. When I let the remote function do its job, the error list gets erased, and then the form is declared 'valid'.

PR #1712, opened 2016-02-14
Core: Unhighlighting field if already highlighted when using remote rule

skjelbek:
Addresses issue #1669, #1375 and #12. Had same problem in our solution.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -606,7 +606,7 @@ $.extend( $.validator, {
 			return $( this.settings.errorElement + "." + errorClass, this.errorContext );
 		},
 
-		resetLists: function() {
+		resetInternals: function() {
 			this.successList = [];
 			this.errorList = [];
 			this.errorMap = {};
@@ -615,7 +615,7 @@ $.extend( $.validator, {
 		},
 
 		reset: function() {
-			this.resetLists();
+			this.resetInternals();
 			this.currentElements = $( [] );
 		},
 
@@ -1427,7 +1427,7 @@ $.extend( $.validator, {
 					validator.settings.messages[ element.name ][ method ] = previous.originalMessage;
 					if ( valid ) {
 						submitted = validator.formSubmitted;
-						validator.resetLists();
+						validator.resetInternals();
 						validator.setToHideFor( element );
 						validator.formSubmitted = submitted;
 						validator.successList.push( element );
--- a/src/core.js
+++ b/src/core.js
@@ -624,9 +624,13 @@ $.extend( $.validator, {
 			this.toHide = this.errors().add( this.containers );
 		},
 
+		setToHideFor: function( element ){
+			this.toHide = this.errorsFor( element );
+		},
+
 		prepareElement: function( element ) {
 			this.reset();
-			this.toHide = this.errorsFor( element );
+			this.setToHideFor( element );
 		},
 
 		elementValue: function( element ) {
@@ -1424,6 +1428,7 @@ $.extend( $.validator, {
 					if ( valid ) {
 						submitted = validator.formSubmitted;
 						validator.resetLists();
+						validator.setToHideFor( element );
 						validator.formSubmitted = submitted;
 						validator.successList.push( element );
 						validator.invalid[ element.name ] = false;
--- a/src/core.js
+++ b/src/core.js
@@ -624,7 +624,7 @@ $.extend( $.validator, {
 			this.toHide = this.errors().add( this.containers );
 		},
 
-		setToHideFor: function( element ){
+		setToHideFor: function( element ) {
 			this.toHide = this.errorsFor( element );
 		},
 
--- a/src/core.js
+++ b/src/core.js
@@ -624,13 +624,9 @@ $.extend( $.validator, {
 			this.toHide = this.errors().add( this.containers );
 		},
 
-		setToHideFor: function( element ) {
-			this.toHide = this.errorsFor( element );
-		},
-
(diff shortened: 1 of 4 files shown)


src/core.js
```diff
@@ -606,12 +606,16 @@ $.extend( $.validator, {
 			return $( this.settings.errorElement + "." + errorClass, this.errorContext );
 		},
 
-		reset: function() {
+		resetLists: function() {
```
staabm: we need a better name for this method.

staabm:
It feels like this change should be manually tested thoroughly.

@suclogger @florianengel @cvwillegen could you verify that this PR fixes the issue you reported?

src/core.js
skjelbek: I fully agree (banged my head against the wall for 20 minutes trying to name this). Any suggestions?

Arkni: What about `resetInternals` ?

staabm: Or just `_reset` ?

Arkni: I think it's a bit confused for someone who started reading the code for the first time as it's the only method that has the name started with an `_`

staabm: Ok, I am fine with resetInternals

test/methods.js
```diff
@@ -618,6 +618,55 @@ test( "remote, highlight all invalid fields", function( assert ) {
...
+	        rules: {
+				something25: {
+	                required: true,
+	                remote: {
+	                    url: "callback.php",
+	                    type: "post",
+						data: {
+							responseText: response
+						},
+						async: false
```
Arkni: The use of `synchronous` calls is discouraged due to the negative effects to the user experience.

test/test.js
```diff
@@ -49,6 +49,15 @@ $.mockjax( {
 } );
 
 $.mockjax( {
+	url: "callback.php",
+	response: function( settings ) {
+		this.responseText = settings.data.responseText();
+		this.responseStatus = settings.data.responseStatus || 200;
+		this.responseTime = settings.data.responseTime || 100;
+	}
+} );
```
Arkni: I see no need in registering another `mockjax` handler, there is already a similar one with url `response.php`, please adapt your test to use it instead.

test/methods.js
```diff
@@ -618,6 +618,55 @@ test( "remote, highlight all invalid fields", function( assert ) {
...
+		$somethingField = $form.find( "input[name='something25']" ),
+		responseText = "false",
+		response = function() { return responseText; },
+		validateOptions = {
+			highlight: function( e ) {
+				$( e ).addClass( "error" );
+			},
+			unhighlight: function( e ) {
+				$( e ).removeClass( "error" );
+			},
```
Arkni: Why overriding `highlight` and `unhighlight` methods ?

Arkni:
I left some comments, I will recheck soon and try to test it locally.

test/methods.js
skjelbek: The async: false was necessary in order to prove that the the test failed before and passed after the change in core.js. The test passes if it's removed, but not for the right reasons. The point of the change in core.js is to avoid resetting this.currentElements inside the success callback before validator.showErrors() is invoked.

test/methods.js
skjelbek: I stumbled upon the bug in our project when our unhighlight method override was not invoked after a successful remote validation if the field was in an invalid state. This test mimicks this behaviour.

test/methods.js
Arkni: Ah, make sense. Thanks for clarification :)

test/methods.js
Arkni: But you have to assert that the `unhighlight` was called by adding an assert (`ok( true, "unhughlight should be called" )` for example) inside `unhighlight`.
WDYT ?

skjelbek: I agree. We still need the existing asserts though, as it is important that the unhighlight is the last one called.

test/test.js
skjelbek: Refactored the response.php handler to support callbacks as well.

src/core.js
```diff
@@ -620,9 +624,13 @@ $.extend( $.validator, {
 			this.toHide = this.errors().add( this.containers );
 		},
 
+		setToHideFor: function( element ) {
```
staabm: I am fine with not adding this new method but leave the 2 call sites (with a duplicated line)

skjelbek: Refactored!

staabm:
@Arkni any feedback. it looks good to me.

Arkni:
LGTM :)

staabm:
@suclogger @florianengel @cvwillegen could you verify that this PR fixes the issue you reported (and does not introduce new issues)?

cvwillegen:
Hi all,

On Wed, Feb 17, 2016 at 4:42 PM, Markus Staab <notifications@github.com> wrote:
> @suclogger @florianengel @cvwillegen could you verify that this
> PR fixes the issue you reported (and does not introduce new issues)?

Sorry, haven't worked at that company for over a year, not sure if
they still use it :-(

(@cvwillegen)

suclogger:
Sure.I would verify it tomorrow morning cause it's midnight here.

suclogger:
It's quite wired that I can't reappear #1669.But this pr works fine to me. @staabm

staabm:
squashed, rebased & merged. thx @skjelbek @Arkni @suclogger for your support.

PR #1717, opened 2016-02-24
Core: `resetForm` should also remove `valid` class from elements

### LINKED ISSUE #1708: `resetForm` should also remove `valid` class from elements
I noticed that calling `validator.resetForm()` on a form that contain some valid elements and others invalid, only removes highlights (`error` class) and hides errors, but the `valid` class still attached to previously valid elements.

IMO, calling `resetForm()` should also remove the `valid` class.

The code responsible for this is the following ([src/core.js#L528-L534](https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L528-L534)):
```js
if ( this.settings.unhighlight ) {
	for ( i = 0; elements[ i ]; i++ ) {
		this.settings.unhighlight.call( this, elements[ i ],
			this.settings.errorClass, "" );
	}
} else {
	elements.removeClass( this.settings.errorClass );
}
```

I guess it should be changed to:
```js
var classes = this.settings.errorClass + " " + this.settings.validClass
if ( this.settings.unhighlight ) {
	for ( i = 0; elements[ i ]; i++ ) {
		this.settings.unhighlight.call( this, elements[ i ], classes, "" );
	}
} else {
	elements.removeClass( classes );
}
```

I didn't create a jsfiddle as it's easy to reproduce using `demo/index.html`.

//Cc @staabm

Arkni:
Closes #1708

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -529,9 +529,12 @@ $.extend( $.validator, {
 				for ( i = 0; elements[ i ]; i++ ) {
 					this.settings.unhighlight.call( this, elements[ i ],
 						this.settings.errorClass, "" );
+					this.findByName( elements[ i ].name ).removeClass( this.settings.validClass );
 				}
 			} else {
-				elements.removeClass( this.settings.errorClass );
+				elements
+					.removeClass( this.settings.errorClass )
+					.removeClass( this.settings.validClass );
 			}
 		},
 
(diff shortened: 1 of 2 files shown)


staabm:
merged via https://github.com/jzaefferer/jquery-validation/commit/f07a267d1e7273f0994c7439a5606721cdf04edb

PR #1731, opened 2016-03-07
fix issue #1730

### LINKED ISSUE #1730: `focusInvalid` focus element failed
there dosn't focus the first visible invalid element on submit or call  `focusInvalid` method manually while the first invalid element is a hidden one.

Alucelx:
Fixes #1730

### DIFF
--- a/dist/jquery.validate.js
+++ b/dist/jquery.validate.js
@@ -1,0 +1,1533 @@
+/*!
+ * jQuery Validation Plugin v1.15.1-pre
+ *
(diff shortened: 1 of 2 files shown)


Arkni:
You should add a test to make sure that it fails without your patch and pass otherwise

staabm:
Please add a test

Alucelx:
@Arkni @staabm 

this is the fixed test

https://jsfiddle.net/Alucelx/480b2osg/5/

PR #1733, opened 2016-03-08
Core: use `input` event with `keyup` as its fallback

### LINKED ISSUE #1715: onclick validation method not only for checkboxes and radio buttons ?
Issue with instant validation. 

When form field is active autosaved values are proposed (default for every form). If we choose autosaved value, by clicking on it, form field is not validated instantly, in other words error message is not removed despite value passes validation.

For example, we have a field "name" and rule of minimum 3 characters.
When we typed one character in the field, and switched to another field, field did not pass instant validation and error is populated. 

If we return to this field and type "Bob" error is removed instantly (seems like http://jqueryvalidation.org/validate#onkeyup method handles this).

But if we return to the field "name" and instead of typing erase everything and select "Bob" from proposed autosaved values, error is not removed, even if we switch to another form field (maybe http://jqueryvalidation.org/validate#onfocusout method should hadle this but it does not).    

This makes confusion in user experience, because user sees that validation is instant and errors are removed immediately after field passes validation. But in this case user selects correct value from autosaved items but error is not removed (field is not validated instantly).

There are workarounds, e.g. disable autosave for form or disable instant validation (set to false all methods except onsubmit). But how to resolve this with active instant validation?

Arkni:
Fixes #1715

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -272,6 +272,16 @@ $.extend( $.validator, {
 				this.element( element );
 			}
 		},
+		oninput: function( element ) {
+			if ( element.name in this.submitted || element.name in this.invalid ) {
+				this.element( element );
+			} else {
+				return;
+			}
+		},
+
+		// DEPRECATED
+		// Please use `oninput` instead.
 		onkeyup: function( element, event ) {
 
 			// Avoid revalidate the field when pressing one of the following keys
@@ -366,6 +376,8 @@ $.extend( $.validator, {
 			this.reset();
 
 			var groups = ( this.groups = {} ),
+				inputSupported = "oninput" in document.createElement( "input" ),
+				inputOrKeyup = ( inputSupported ? "input" : "keyup" ) + ".validate",
 				rules;
 			$.each( this.settings.groups, function( key, value ) {
 				if ( typeof value === "string" ) {
@@ -390,7 +402,7 @@ $.extend( $.validator, {
 			}
 
 			$( this.currentForm )
-				.on( "focusin.validate focusout.validate keyup.validate",
+				.on( "focusin.validate focusout.validate " + inputOrKeyup,
 					":text, [type='password'], [type='file'], select, textarea, [type='number'], [type='search'], " +
 					"[type='tel'], [type='url'], [type='email'], [type='datetime'], [type='date'], [type='month'], " +
 					"[type='week'], [type='time'], [type='datetime-local'], [type='range'], [type='color'], " +
--- a/src/core.js
+++ b/src/core.js
@@ -390,6 +390,10 @@ $.extend( $.validator, {
 				rules[ key ] = $.validator.normalizeRule( value );
 			} );
 
+			// Delete `onkeyup` if `input` event is supported
+			// Otherwise, remove `oninput` handler
+			delete this.settings[ inputSupported ? "onkeyup" : "oninput" ];
+
 			function delegate( event ) {
 				var validator = $.data( this.form, "validator" ),
 					eventType = "on" + event.type.replace( /^validate/, "" ),
--- a/src/core.js
+++ b/src/core.js
@@ -375,7 +375,7 @@ $.extend( $.validator, {
 
 			var groups = ( this.groups = {} ),
 				inputSupported = "oninput" in document.createElement( "input" ),
-				inputOrKeyup = ( inputSupported ? "input" : "keyup" ) + ".validate",
+				inputOrKeyup = inputSupported ? "input" : "keyup",
 				rules;
 			$.each( this.settings.groups, function( key, value ) {
 				if ( typeof value === "string" ) {
@@ -404,7 +404,7 @@ $.extend( $.validator, {
 			}
 
 			$( this.currentForm )
-				.on( "focusin.validate focusout.validate " + inputOrKeyup,
+				.on( "focusin.validate focusout.validate " + inputOrKeyup + ".validate",
 					":text, [type='password'], [type='file'], select, textarea, [type='number'], [type='search'], " +
 					"[type='tel'], [type='url'], [type='email'], [type='datetime'], [type='date'], [type='month'], " +
 					"[type='week'], [type='time'], [type='datetime-local'], [type='range'], [type='color'], " +
--- a/src/core.js
+++ b/src/core.js
@@ -275,8 +275,6 @@ $.extend( $.validator, {
 		oninput: function( element ) {
 			if ( element.name in this.submitted || element.name in this.invalid ) {
 				this.element( element );
-			} else {
-				return;
 			}
 		},
 
--- a/test/.jshintrc
+++ b/test/.jshintrc
@@ -24,6 +24,8 @@
 		"equal": false,
 		"expect": false,
 		"Globalize": false,
+		"inputOrKeyup": true,
+		"inputSupported": true,
 		"methodTest": false,
 		"module": false,
 		"notEqual": false,
@@ -35,7 +37,6 @@
 		"strictEqual": false,
 		"stop": false,
 		"test": false,
-		"TestHelpers": true,
 		"throws": false,
 		"JSHINT": false
 	}
--- a/test/error-placement.js
+++ b/test/error-placement.js
@@ -162,10 +162,10 @@ test( "errorcontainer, show/hide only on submit", function() {
 	ok( labelContainer.is( ":visible" ), "must be visible" );
 	ok( container.is( ":visible" ), "must be visible" );
 
-	$( "#firstname" ).val( "hix" ).keyup();
-	$( "#testForm1" ).triggerHandler( "keyup", [
+	$( "#firstname" ).val( "hix" ).trigger( inputOrKeyup );
+	$( "#testForm1" ).triggerHandler( inputOrKeyup, [
 			jQuery.event.fix( {
-				type: "keyup",
+				type: inputOrKeyup,
 				target: $( "#firstname" )[ 0 ]
 			} )
 		] );
@@ -313,16 +313,16 @@ test( "test aria-describedby with input names contains CSS-selector meta-charact
 
 	// Re-run validation
 	field.val( "some" );
-	field.trigger( "keyup" );
+	field.trigger( inputOrKeyup );
 
 	field.val( "something" );
-	field.trigger( "keyup" );
+	field.trigger( inputOrKeyup );
 
 	equal( field.attr( "aria-describedby" ), "testForm21!#$%&'()*+,./:;<=>?@[\\]^`{|}~-error", "`aria-describedby` should remain the same as before." );
 
 	// Re-run validation
 	field.val( "something something" );
-	field.trigger( "keyup" );
+	field.trigger( inputOrKeyup );
 
 	ok( field.valid() );
 	equal( field.attr( "aria-describedby" ), "testForm21!#$%&'()*+,./:;<=>?@[\\]^`{|}~-error", "`aria-describedby` should remain the same as before." );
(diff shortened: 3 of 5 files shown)


src/core.js
```diff
@@ -272,6 +272,16 @@ $.extend( $.validator, {
...
+		oninput: function( element ) {
+			if ( element.name in this.submitted || element.name in this.invalid ) {
+				this.element( element );
+			} else {
+				return;
+			}
+		},
+
+		// DEPRECATED
+		// Please use `oninput` instead.
```
Arkni: I'm not the one who should take the decision, but I think it's ok to deprecate `onkeyup` and use `oninput` instead.
If this get deprecated, then we can remove it in `2.0.0` which should also define the supported browsers (`>IE9`).

test/.jshintrc
```diff
@@ -24,6 +24,8 @@
 		"equal": false,
 		"expect": false,
 		"Globalize": false,
+		"inputOrKeyup": true,
```
staabm: Those 2 vars are not global...?

src/core.js
```diff
@@ -272,6 +272,16 @@ $.extend( $.validator, {
 				this.element( element );
 			}
 		},
+		oninput: function( element ) {
+			if ( element.name in this.submitted || element.name in this.invalid ) {
```
staabm: Do we need the empy val check like we have it in onkeyup?

src/core.js
```diff
@@ -272,6 +272,16 @@ $.extend( $.validator, {
...
+				this.element( element );
+			} else {
+				return;
+			}
+		},
+
+		// DEPRECATED
+		// Please use `oninput` instead.
 		onkeyup: function( element, event ) {
```
staabm: Would it make sense to not register onkeyup at all when oninput is supported?

test/error-placement.js
```diff
@@ -162,10 +162,10 @@ test( "errorcontainer, show/hide only on submit", function() {
 	ok( labelContainer.is( ":visible" ), "must be visible" );
 	ok( container.is( ":visible" ), "must be visible" );
 
-	$( "#firstname" ).val( "hix" ).keyup();
```
staabm: Wdyt think about leaving keyup tests as is and add additional oninput tests.. That we we can more easily delete keyup coverage and we lso make sure that onkeyup handling works in browsers which support both

src/core.js
```diff
@@ -272,6 +272,16 @@ $.extend( $.validator, {
 				this.element( element );
 			}
 		},
+		oninput: function( element ) {
+			if ( element.name in this.submitted || element.name in this.invalid ) {
+				this.element( element );
+			} else {
+				return;
```
staabm: Why do we need this empty else block?

src/core.js
Arkni: Make sense.
I'll see what I can do.

test/error-placement.js
Arkni: I agree, but you have to know that I changed the attachment of the event in `init()` to use `input` event instead of `keyup` if the former is supported, and to use the latter otherwise. Should I change the logic ?

src/core.js
Arkni: Ah, I forgot to delete when making the changes. I'll omit it.

src/core.js
Arkni: `input` event fires only when the value changes. The empty val check in `onkeyup` was added to prevent validating empty element when losing focus. So I don't see a need for it here.

But I'm open to suggestion.

staabm: Will it not fire a evenr when changing from a filled field to an empty one?

test/error-placement.js
staabm: Sounds like we should leave it as is, thx for explaining

src/core.js
Arkni: Nope, the `keyup` event will be fired then not the `input` event.

src/core.js
Arkni: Would it be ok to remove one of the 2 handlers if one of the events is supported.
So:
* if `input` event is supported, we can then delete `onkeyup`
* else, remove `oninput`.

So the user now is supposed to use the 2 handlers to support legacy browsers (even though I think it's a waste of time trying to support <IE9 and other legacy browsers)

staabm: You mentioned that input is not invoked when field value gets empty. Does it work without keyup at all?

Arkni: In the usecase you mentioned above, the events that will fire are the `blur` (on first element) and `keyup` (on second element).
Input event is fired synchronously when the value of an element is changed.

> Does the lib work without keyup at all?

Yes and without any problem (as far as I can tell with the manually tests I did myself)

Arkni: But as you know, removing `onkeyup` is a BC break and should be done in version `2.0.0`

staabm: Let me re-phrase my question: can we get rid of keyup handler in browsers which support the input event?

When exactly will the input event be triggered? On each keystroke?

Arkni: Sorry if I misunderstood your question.

> can we get rid of keyup handler in browsers which support the input event?

Yes we can.

> When exactly will the input event be triggered? On each keystroke?

Allow me to point to specs, see:
* https://developer.mozilla.org/en-US/docs/Web/Events/input
* https://w3c.github.io/editing/input-events.html

staabm: Thx for the links.

So as I understand it, this will also add validation e.g. after paste content into a input. We have to test carefully.

> if input event is supported, we can then delete onkeyup
else, remove oninput.

Ok.

> So the user now is supposed to use the 2 handlers to support legacy browsers (even though I think it's a waste of time trying to support <IE9 and other legacy browsers)

We need to document that

Arkni: > So as I understand it, this will also add validation e.g. after paste content into a input. We have to test carefully.

Correct.

>> So the user now is supposed to use the 2 handlers to support legacy browsers (even though I think it's a waste of time trying to support <IE9 and other legacy browsers)

> We need to document that

I'll send a PR to document the `oninput` handler asap, and sure I'll mention this.

staabm: > I'll send a PR to document the oninput handler asap, and sure I'll mention this.

First we should make sure the implementations works as expected in all supported browsers without side-effects. The docs can be done later on.

Arkni: I'll try to do my best on my side.

staabm: Thank you so much :tada:

src/core.js
```diff
@@ -380,6 +390,10 @@ $.extend( $.validator, {
 				rules[ key ] = $.validator.normalizeRule( value );
 			} );
 
+			// Delete `onkeyup` if `input` event is supported
+			// Otherwise, remove `oninput` handler
+			delete this.settings[ inputSupported ? "onkeyup" : "oninput" ];
+
```
Arkni: As discussed above, one of the 2 handlers (`onkeyup` and `oninput` should be removed based on `input` event support).

src/core.js
```diff
@@ -366,6 +374,8 @@ $.extend( $.validator, {
 			this.reset();
 
 			var groups = ( this.groups = {} ),
+				inputSupported = "oninput" in document.createElement( "input" ),
+				inputOrKeyup = ( inputSupported ? "input" : "keyup" ) + ".validate",
```
staabm: Per varname I would move the string concat to the place we use it

staabm:
Looks good so far... Will do some testing in real browsers in the next days

src/core.js
Arkni: Fixed in 7147b81

Arkni:
I'll squash the commits once the manually tests confirm that this implementation works as expected.

staabm:
@ultraviolet could you confirm this fixes your issue?

ultraviolet:
> @ultraviolet could you confirm this fixes your issue (and doesnt introduce new ones)?
I can only test already built version of plugin (which is available for download). As of time of writing it is 1.15.0 which i got from plugin's site. 
I suppose it does not have modifications discussed in this branch yet.
Nevertheless, i tested v1.15.0 (just in case) and plugin's behavior is the same i described previously. 

I also included script "additional-methods.js", though i have no idea what is its purpose (in the beginning of its code i didn't find info in comments, neither anywhere on site).

staabm:
@ultraviolet please try this https://gist.githubusercontent.com/staabm/240fa96929f0429f6fb2/raw/9b3eae294fbcf360f380c161c14ef9f7dda4162a/jquery.validate.js

ultraviolet:
> https://gist.githubusercontent.com/staabm/240fa96929f0429f6fb2/raw/9b3eae294fbcf360f380c161c14ef9f7dda4162a/jquery.validate.js

Works like a charm - that issue i described is gone with this build (validation is instant when choosing form's saved string).

However, some new issue appeared (which does not exist with production version):
Phone number length is not validated even if entered correctly (error remains). But if form is sent, it is validated successfully. So, error is misleading and should dissapear right after length is correct, but it's not.

Masked input of field provided by plugin http://digitalbush.com/projects/masked-input-plugin/ : 
```
$(".phoneField").mask("?+7 (999) 999-99-99");
```

jQuery validation code for phone field:
 ```
	jQuery.validator.addMethod("phoneLength", function (value, element) {
		var phone_value = value.replace(/\D+/g,"");
		return (phone_value.length == 11);
	}, "Too short number");
 
	jQuery('#superForm').validate({
		onfocusout: false
        }		
	}); 
 
	jQuery.validator.addClassRules("phone", {
		phoneLength: true
	});	
```

Steps to reproduce: fill out phone field incompletely (jQuery validation plugin is silent now because of "onfocusout: false"), try to send form. Error about phone length is populated, then fill out phone field correctly.

Result: error "Too short number" remains.

Expected result (which i get with production version of plugin): error should dissapear immediately after phone length is correct.

**Comment by Arkni** [comment]
Thanks @ultraviolet 

Will investigate the issue when time allows.

**Comment by Arkni** [comment]
Hi @ultraviolet 

I did some testing and I found that [`jQuery masked input`](http://digitalbush.com/projects/masked-input-plugin/) prevents `input` event from bubbling (in fact, it prevents `keypress` event from bubbling which lead to `input` event not firing) which is bad as a design, _plugins should never stop events from propagating_.

Also, to fix your issue (#1715) we have to use the `input` event too (in addition to `keyup` event). But then we will be found with two events handlers `keyup` and `input` to listen to element changes, so here is how we proceed:
* Keep the later and drop the former in modern browsers.
* Keep the former (`keyup`) and drop the later (`input`) in legacy browsers.

So currently, you have to use the current version (`1.15.0`) and the plugin you mentioned. In this case, to enable the validation on `input` (for copy paste in other elements), you can add the current code:
```js
// Initialize the plugin
var v = $( "your_form_seletor" ).validate( validation_options );

// validate on `input` event
$( "your_element_selector" ).on( "input", function( e ) {
    if ( this.name in v.submitted || this.name in v.invalid ) {
        v.element(this);
    }
} );
```

Hmm, I guess this PR should remain open for now, If people can add validation on `input` event easily, then we can keep the plugin as is without adding support for `input` event.

What do you guys think ?

staabm:
> Hmm, I guess this PR should remain open for now, If people can add validation on input event easily, then we can keep the plugin as is without adding support for input event.

doesnt sound too bad. we should mention this into the docs how one can validate this on-autocomple then.

Arkni:
> doesnt sound too bad. we should mention this into the docs how one can validate this on-autocomple then.

Yes. I'll try to update the docs when I get a chance.
Should we close this PR then ?

staabm:
Yeah, maybe we can get this in a more native fashion in 2.0

ultraviolet:
> So currently, you have to use the current version (1.15.0) and the plugin you mentioned. In this case, to enable the validation on input (for copy paste in other elements), you can add the current code:

Thanks a lot! This is excellent workaround and can be easily implemented.

PR #1734, opened 2016-03-09
min rule

DhruvProtocol:
If the filed is allowed to be number then it accepts comma in value (i.e. 1200 and 1,200 both are valid). Now if "min" rule is provided with number then currently even if i'm entering 1,200 error(Please enter a value greater than or equal to 0.) is throwing. So if the current element has number and min both rules then we need to replace comma with 0 length string.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1369,6 +1369,8 @@ $.extend( $.validator, {
 
 		// http://jqueryvalidation.org/min-method/
 		min: function( value, element, param ) {
+			if (this.settings.rules[element.name].number)
+				value = value.replace(/,/g, '');
 			return this.optional( element ) || value >= param;
 		},
 


Arkni:
The Travis check has failed (https://travis-ci.org/jzaefferer/jquery-validation/builds/114806911)

staabm:
Depending on local the `,` can be a thousands-seperator, therefore this cannot be merged as is. Looks like something which should be added with Globalize support.

Arkni:
I'm going to close this as it's not something this plugin has to support. As @staabm said in his comment, the `,` depends on the local and how numbers are formated, not how they are entered. What I can suggest people to do is to use the [`normalizer`](https://jqueryvalidation.org/normalizer/) to remove the `,` before hitting the validation:
```js
$("form").validate({
  rules: {
    field: {
      required: true,
      min: 1500,
      normalizer: function( value ) {
        return value.replace(/,/g, "");
      }
    }
  }
});
```

PR #1741, opened 2016-03-11
add validate method for positive integers

KarmaLaBelle:
I had an assignment that required this, saw it wasn't a thing, so wanted to go ahead and make a pull request.

http://codepen.io/anon/pen/jqrpPx?editors=1010

New to the whole Pull Request thing, so if you have any input, let me know! :smiley:


staabm:
I dont think this kind of method needs to be part of the core. Its very easy to build on app level and is only required for some apps, not most.

Thanks for taking the time to contribute

Issue #1745, opened 2016-03-15
Can't remove 'required'-rule when element has 'required'-attribute


ischas:
When the element has a 'required'-attribute, you can't remove it by calling `$(element).rules('remove','required')`.

https://jsfiddle.net/6hnw855z/1/

A pull-request to solve the problem will follow up.

k0stenko:
try to use double quotes with "remove"
$(element).rules("remove",'required')

Arkni:
Closing for now. See https://github.com/jquery-validation/jquery-validation/pull/1746#issuecomment-372032944 for reason.

PR #1746, opened 2016-03-15
remove "required"-attribute from element when removing "required"-rule

ischas:
Since the "attributeRules"-method checks for the "required"-attribute of the element, this attribute has to be removed when removing the "required"-rule.

This addresses #1745.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -174,6 +174,7 @@ $.extend( $.fn, {
 			delete data.required;
 			data = $.extend( { required: param }, data );
 			$( element ).attr( "aria-required", "true" );
+			$( element ).attr( "required", "true" );
 		}
 
 		// Make sure remote is at back
--- a/src/core.js
+++ b/src/core.js
@@ -409,6 +409,10 @@ $.extend( $.validator, {
 			// Add aria-required to any Static/Data/Class required fields before first validation
 			// Screen readers require this attribute to be present before the initial submission http://www.w3.org/TR/WCAG-TECHS/ARIA2.html
 			$( this.currentForm ).find( "[required], [data-rule-required], .required" ).attr( "aria-required", "true" );
+			
+			// Add required to any Static/Data/Class required fields before first validation
+			// Required is the html-equivalent to aria-required https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute
+			$( this.currentForm ).find( "[aria-required], [data-rule-required], .required" ).attr( "required", "true" );
 		},
 
 		// http://jqueryvalidation.org/Validator.form/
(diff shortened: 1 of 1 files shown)


staabm:
I am not sure we should do this. since the lib did not add this required attribute, we shouldnt remove it IMO.

to be honest I think you should add the required rule using `$(element).rules('add','required')` when you rely on `$(element).rules('remove','required')` to remove it again.

ischas:
You're right when you say that my supposed change isn't symmetric. I'd suggest to change my pull request so the required-attribute is added when adding the required-rule.

Then the behaviour would be equal to the "aria-required"-attribute.

Arkni:
From my point of view, this lib shouldn't remove attributes added by the user.

>to be honest I think you should add the required rule using $(element).rules('add','required') when you rely on $(element).rules('remove','required') to remove it again.

:+1: to this.

Arkni:
> I changed my pull request so the required-attribute is added when adding the required-rule.
> 
> Now the behaviour is equal to the "aria-required"-attribute.

In that case, you should fix the travis failure: https://travis-ci.org/jzaefferer/jquery-validation/builds/116632270#L193

staabm:
> You're right when you say that my supposed change isn't symmetric. I changed my pull request so the required-attribute is added when adding the required-rule.

hehe, I didnt foresee that you would just add the attribute... ;-)

I am still not convinced we should add/remove the `required` attribute.
per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute ´required` and `aria-required` are identical in a html5 document.

In previous html versions there is no "required" attribute defined.

I have to think about this change in general..

staabm:
@ischas I didnt meant you need to adjust this PR furhter... I have to think about the general idea of this change.

wouldnt it be possible for you to use `$(element).rules('add','required')` instead of rendering a `required` attribute? Is there a reason why you dont do that?

ischas:
We are collaborating with a third company, which generates the HTML. In some circumstances the validation rules of fields change and our JavaScript is responsible for this (client-side) changes.

If you prefer not to merge my suggested changes I have to remove the required-attribute in my code.

If you change your mind you must help me to get travis happy. ;-)

ischas:
> From my point of view, this lib shouldn't remove attributes added by the user.
In fact, at the moment the "aria-required"-attributes is removed by code even it was added by the user.

https://jsfiddle.net/ischas/6hnw855z/2/

Arkni:
You're right. I didn't take that into consideration because this lib will add that attribute to a required element to make it possible for `Screen readers` to announce the element as `required`. And delete it when the `required` rule is removed by the user.

Arkni:
We dropped support for `aria-required` attribute in #2012. And I still stand by the point made by @staabm in https://github.com/jquery-validation/jquery-validation/pull/1746#issuecomment-197790882 and my own that this lib should not remove attributes added by the developer.

As a result, if anyone want to remove the `required` rule alongside the `required` attribute, they should do so by doing:
```js
$(element).rules('remove', 'required');
$(element).removeAttr('required');
```

I will close this PR for now. Feel free to ask if you have followup questions.
Thanks for oppening this PR, though.

PR #1755, opened 2016-03-22
No need for spaces inside brackets

hugovk:
In English, it looks a bit odd with spaces inside brackets.


Arkni:
Would you mind editing this line too: https://github.com/jzaefferer/jquery-validation/blob/3de30ac8a173c844fab128ef0ebdef59a6b4e964/src/localization/messages_ge.js#L12

hugovk:
@Arkni Done!

staabm:
thx for the PR. did you check all localization files for this "problem"?

hugovk:
Those are the only two like that.

staabm:
thx

PR #1765, opened 2016-04-12
If comparing target is not found, throw an error.

gqqnbig:
If comparing target is not found, throw an error.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1402,6 +1402,8 @@ $.extend( $.validator, {
 		equalTo: function( value, element, param ) {
 
 			// Bind to the blur event of the target in order to revalidate whenever the target field is updated
+			if (target.size() === 0)
+				throw new Error("Comparing target is not found.");
 			var target = $( param );
 			if ( this.settings.onfocusout && target.not( ".validate-equalTo-blur" ).length ) {
 				target.addClass( "validate-equalTo-blur" ).on( "blur.validate-equalTo", function() {


src/core.js
```diff
@@ -1402,6 +1402,8 @@ $.extend( $.validator, {
 		equalTo: function( value, element, param ) {
 
 			// Bind to the blur event of the target in order to revalidate whenever the target field is updated
+			if (target.size() === 0)
```
staabm: Target is defined in the lines below.. This cannot work

staabm:
Why do you think this change is required? What are you trying to achieve?

gqqnbig:
@staabm if the comparing target is not found, this control cannot have any text.  If it's also required, validations always fail. That the comparing target is not found often indicates bad coding that programmer forgets to unset equal-to attribute, or forgets to show the comparing target.

It's my bad that writing it in wrong line..

staabm:
AFAIR its "jquery-style" to not error on empty selectors

Arkni:
If `target` is not found then `equalTo` method will always return `false`, and this itself an indication that the programmer did something wrong (wrong value or a debug is needed to make sure everything work as expected).
But throwing an exception will save the programmer the time of debugging and will help him|her fixing the problem very quickly. So, I'm OK with this change.

The error msg can be improved to include the target selector (`param`).

staabm:
Ok. Lets land it after we have a unit test and the message is adjusted

PR #1771, opened 2016-04-19
core: Fix error when calling $.fn.rules() on a non-form element.

### LINKED ISSUE #1770: $.fn.rules throws an error when called on a jQuery object whose first element is not a form element.
Calling $.`fn.rules()` throws an error when called on a jQuery object whose first element is not a form element, i.e. has a `.form` property.

version 1.15.0

This is similar to #1706. Changes for that issue involved returning `undefined` when calling `$.fn.rules()` on an empty jQuery object.

I'll create a pull request to handle this case, but I think it would make more sense for this case and the case from #1706 to return an empty object, rather than `undefined`, seeing as the return value on a successful call is an object containing validation rules.  I can certainly change this back to returning `undefined` if you disagree.

jsclark:
Closes #1770

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -117,14 +117,13 @@ $.extend( $.fn, {
 
 	// http://jqueryvalidation.org/rules/
 	rules: function( command, argument ) {
-
-		// If nothing is selected, return nothing; can't chain anyway
-		if ( !this.length ) {
-			return;
-		}
-
 		var element = this[ 0 ],
 			settings, staticRules, existingRules, data, param, filtered;
+
+		// If nothing is selected, return empty object; can't chain anyway
+		if ( element == null || element.form == null ) {
+			return {};
+		}
 
 		if ( command ) {
 			settings = $.data( element.form, "validator" ).settings;
--- a/test/rules.js
+++ b/test/rules.js
@@ -397,9 +397,14 @@ test( "rules(), normalizer", function() {
 	}, "This should throw a TypeError exception." );
 } );
 
-test( "rules() - on empty jquery set #1706", function() {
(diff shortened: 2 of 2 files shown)


src/core.js
```diff
@@ -117,15 +117,14 @@ $.extend( $.fn, {
...
-		if ( !this.length ) {
-			return;
-		}
-
 		var element = this[ 0 ],
 			settings, staticRules, existingRules, data, param, filtered;
 
+		// If nothing is selected, return empty object; can't chain anyway
+		if ( element == null || element.form == null ) {
+			return {};
```
staabm: Why do you opt for a empty object?

staabm:
Please add a unit test

src/core.js
jsclark: I went with an empty object because this method normally returns an object.  I called this out in my bug report (#1770), and I don't feel strongly about `{}` over `undefined`, but I think keeping the return type consistent is reasonable.

jsclark:
@staabm I modified one test and added another, is that not sufficient?  I'm happy to add more tests, but I need clarification on what else you're looking for.

src/core.js
staabm: I would prefer the intial return value

staabm:
Tests are looking good. Did you verify they fail without your patch?

It doesnt matter how many commit you add. I will take care of it on merge.

jsclark:
Yes, I confirmed that they fail without my changes to core.js.  I'll need to modify the when I change the return value from `{}` back to `undefined` though.  I'll make that change tonight or over the weekend.

staabm:
Adjusted the last few bits and rebased, merged it. Thanks @jsclark

Issue #1681, opened 2016-01-11
Update qunit.


staabm:
`asyncTest` is deprecated and will be removed in the next major version of `Qunit`, from [Docs](http://api.qunitjs.com/QUnit.asyncTest/):
>This method is deprecated and will be removed in QUnit 2.0.0. Use QUnit.test() in conjunction with assert.async() instead.

See http://qunitjs.com/upgrade-guide-2.x/

PR #1777, opened 2016-05-05
Tests: Upgrade QUnit to 2.0

Arkni:
Ref #1681


package.json
```diff
@@ -37,17 +37,18 @@
...
-		"grunt-contrib-copy": "^0.5",
-		"grunt-contrib-jshint": "^0.11.3",
-		"grunt-contrib-qunit": "^0.4",
-		"grunt-contrib-uglify": "^0.4",
-		"grunt-contrib-watch": "^0.6",
-		"grunt-jscs": "^2.2",
-		"grunt-text-replace": "^0.3.11"
+		"commitplease": "2.3.1",
+		"grunt": "1.0.1",
+		"grunt-contrib-compress": "1.2.0",
```
staabm: Why did you add this hardcoded constraints?

staabm:
I like what I see so far

package.json
Arkni: I followed the convention used in most of `jQuery` projects (`QUnit`, `jquery`, `jquery-ui` ...).

Here is a quote from one of the core developpers of `jQuery`:
>Most of jQuery repos went in the opposite direction. Semver is an idea but reality is that from time to time a package breaks and then we have a problem. IMO it's better to just pin versions and update from time to time. Note that using non-latest version of a dependency isn't a bad thing; maybe they introduced a new feature we don't use, maybe a bug fix that doesn't affect us that much. It's good to not lock down to something ancient but it doesn't mean constantly living on the bleeding edge is an adequate answer either

But if you want, I can leverage all the dev-dependencies.

staabm: As those are only devDeps I can live with whatever you use. Just curious

Arkni:
I'll update this PR as soon as `QUnit@2.0` is released. Currently, only the `rc1` is available and I did this change to make sure everything works a expected with the new major version.

Arkni:
@staabm I think this is good to go.

staabm:
@Arkni I will create a 1.15.1 release now and we can land this PR after that. could you rebase it then?

may I ask you to create a changelog from the current master?

Arkni:
> I will create a 1.15.1 release now and we can land this PR after that. could you rebase it then?

Sure, I will do that.

> may I ask you to create a changelog from the current master?

Of course, just a few minutes.

Arkni:
@staabm 
Here you go: https://gist.github.com/Arkni/8775e2f0755659b98d38ca05b96fa798#file-jquery-validate-1-15-1-md

Cleaned it up a little :)

Arkni:
And PR rebased as requested :)

staabm:
@Arkni thank you so much for your assistance. couldnt have build this release without your support!!

Arkni:
Any time :)

PR #1778, opened 2016-05-05
Core: Adjust `delegate` to work with `contenteditable` elements

Arkni:
I can add more tests if they are needed.


Arkni:
Closed in favor of #1785

PR #1780, opened 2016-05-05
Core: Call `errorPlacement` in validator scope

### LINKED ISSUE #1542: ErrorPlacement is called without validator scope
Unlike all the other functions (highlight, unhighlight and showErrors), errorPlacement is called without the scope of the activate validator.

https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L786
```javascript
this.settings.errorPlacement( place, $( element ) );
```
Suggest this is changed to:
```javascript
this.settings.errorPlacement.call( this, place, $( element ) );
```

Again this is because I'm trying to crowbar in additional functionality, but there are very few locations which allow me to do this.

Is there any reason the scope isn't given to errorPlacement? I can't see one besides "you're not supposed to mess with the internals from here".

Arkni:
Fixes #1542

Do I need to add a test for this?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -905,7 +905,7 @@ $.extend( $.validator, {
 				if ( this.labelContainer.length ) {
 					this.labelContainer.append( place );
 				} else if ( this.settings.errorPlacement ) {
-					this.settings.errorPlacement( place, $( element ) );
+					this.settings.errorPlacement.call( this, place, $( element ) );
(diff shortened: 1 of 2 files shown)


staabm:
Looks good. Test would be awesome

Arkni:
Added the test

staabm:
merged thx.

PR #1781, opened 2016-05-06
validation loosing focus bug

phiggis:
if you click a bunch of times on a drop down, and sometimes click out of
the dropdown (while its showing) the error you will get is the .call on
the rule will be null as the dropdown is the item that no longer has
focus and therefor you will get a javascript error


phiggis:
Sorry this is my first ever bug fix, so I am trying to figure out how to make the CI build pass so I have to figure out how to build this thing locally (which I dont know how to yet as I havent used npm and other stuff)

PR #1785, opened 2016-05-18
Core: Fixed issue with contenteditable elements in forms where events would cause exceptions

svrx:
Where content editable fields are in a <form> exception will happen. 
This is because form validation process hasn't been triggered, but events on the inputs will trigger their validation.

The proposed fix addresses this issue by checking if the elements have been the <form> reference, if they don't and are contenteditable element it will be created at that time.

The solution is the same as in $.validator.elements(), but covers another flow for validating a single input field.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -381,6 +381,10 @@ $.extend( $.validator, {
 			} );
 
 			function delegate( event ) {
+				if (!this.form && this.hasAttribute("contenteditable")) {
+			        this.form = $(this).closest("form")[0];
+			    }
+			    
 				var validator = $.data( this.form, "validator" ),
 					eventType = "on" + event.type.replace( /^validate/, "" ),
 					settings = validator.settings;
--- a/src/core.js
+++ b/src/core.js
@@ -383,7 +383,7 @@ $.extend( $.validator, {
 			function delegate( event ) {
 
 				// Set form expando on contenteditable
-				if ( this.hasAttribute( "contenteditable" ) ) {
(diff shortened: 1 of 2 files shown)


Arkni:
That issue is already fixed in #1778 (not merged yet)

svrx:
@Arkni thanks for the information. 
Just had a look at your suggested fix. 
I believe mine addresses the issue in a better way, as you only pay the penalty of traversing to the parent form on the first time an event is raised in an input, instead of every time for every event.

What do you think? 

Just noticed the build failed for code style issues, if needed can resubmit the pull request with style fixed.

Arkni:
Yeah, good catch!
I'll close mine and point back to this PR.

> if needed can resubmit the pull request with style fixed.

Please make sure to include a test as well.

svrx:
Fixed style issue and added unit test for this issue. 
Any additional comment just let me know.

src/core.js
```diff
@@ -381,6 +381,12 @@ $.extend( $.validator, {
 			} );
 
 			function delegate( event ) {
+
+				// Set form expando on contenteditable
+				if ( this.hasAttribute( "contenteditable" ) ) {
```
Arkni: Shouldn't this be:
```js
if ( !this.form && this.hasAttribute( "contenteditable" ) ) {
```
as it was in your first commit ?

svrx: Thanks for that, you're absolutely right, this seems to have been lost along the process. 
Added it back and should be good to go.

Arkni:
Looks good to me :+1:

staabm:
merged, thx.

PR #1793, opened 2016-05-31
Correct notEqualTo jQuery reference.

TrueWill:
Additional method was referencing the global jQuery object instead of the injected $.

Fixes 1773.


staabm:
Merged manually via https://github.com/jzaefferer/jquery-validation/commit/3c323005e1b2e68877695f3a30c3427b1eb8a554

PR #1801, opened 2016-06-15
[refactor] Regex for E-Mail Address

furkanpoyraz:
This was valid with the previous regex: "test@mail"

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1321,7 +1321,7 @@ $.extend( $.validator, {
 			// Retrieved 2014-01-14
 			// If you have a problem with this implementation, report a bug against the above spec
 			// Or use custom methods to implement your own email validation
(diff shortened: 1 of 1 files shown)


staabm:
See the README.md for details.

> **IMPORTANT NOTE ABOUT EMAIL VALIDATION**. As of version 1.12.0 this plugin is using the same regular expression that the [HTML5 specification suggests for browsers to use](https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address). We will follow their lead and use the same check. If you think the specification is wrong, please report the issue to them. If you have different requirements, consider [using a custom method](http://jqueryvalidation.org/jQuery.validator.addMethod/).
In case you need to adjust the built-in validation regular expression patterns, please [follow the documentation](http://jqueryvalidation.org/jQuery.validator.methods/).

PR #1807, opened 2016-06-27
Added greaterThan and lessThan validations

djhvscf:
#### Description
Added greaterThan and lessThan validations.


staabm:
Does min/max not work for you?

http://jqueryvalidation.org/min-method/
http://jqueryvalidation.org/max-method/

staabm:
will close for now, feel free to discuss it furhter. we can re-open when required.

djhvscf:
it is something like that, but in the min / max methods it is using greaterThanOrEqual, I'm using the greaterThan not Equal

PR #1809, opened 2016-06-28
Update step validation to handle floating points correctly

tbjers:
This is a new PR based on #1760. I couldn't find a way to modify that actual PR with my additional changes so I created a new PR that includes unit tests. @adhayward @staabm 

Tests: Added regression unit tests for PR #1760
Testing small fraction steps to ensure correct modulo operation.
Cleaned up @adhayward's formatting, was using spaces instead of tabs.
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Remainder_()
See: http://wiki.ecmascript.org/doku.php?id=strawman:modulo_operator
See: http://stackoverflow.com/questions/3966484/floating-point-numbers-and-javascript-modulus-operator

PR #1810, opened 2016-06-30
Core: Fix when adding a step of 0.01 that only bits are allowed

DrowningElysium:
When using a step of 0.01 the validation starts to do funky.

When you have a step of 0.01 the following are allowed:
0, 0.01, 0.02, 0.04, 0.08, 0.16, 0.32, 0.64, 1.28, etc...

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1400,7 +1400,7 @@ $.extend( $.validator, {
 			if ( notSupported ) {
 				throw new Error( errorMessage );
 			}
-			return this.optional( element ) || ( value % param === 0 );
+			return this.optional( element ) || ( ( value / param ) % 1 === 0 );
 		},
 
 		// http://jqueryvalidation.org/equalTo-method/


Arkni:
Duplicate of #1760

staabm:
the fix itself looks way more elegant then #1760 .

if covered by a unit test we should merge this one IMO.

staabm:
could you pick the unit tests from https://github.com/jzaefferer/jquery-validation/pull/1809 into this one?

staabm:
@tbjers would you be ok, with this fix instead of yours?

DrowningElysium:
@staabm this fix isn't a fix, check this: https://github.com/jzaefferer/jquery-validation/pull/1760#discussion_r59124104

Did my research to late :/
#1809 might be better, haven't tested it

tbjers:
Researching both right now and will let you know. Need to run the unit tests on this first.

tbjers:
@staabm This PR does not address the original issue, stepping fractions. The problem is with the JavaScript floating point calculations. The native JavaScript `%` operator isn't modulo, it's just remainder. Therefore #1809 correctly addresses the problem. Thanks, @DrowningElysium.

staabm:
@tbjers thx for checkin.

PR #1812, opened 2016-07-07
Fix email validation regex pattern

AlaeddineMessadi:
Email validation was wrong. it's used to accepts "xxx@xx"


nebrot:
See https://github.com/jzaefferer/jquery-validation#reporting-issues-and-contributing-code

> IMPORTANT NOTE ABOUT EMAIL VALIDATION. As of version 1.12.0 this plugin is using the same regular expression that the HTML5 specification suggests for browsers to use. We will follow their lead and use the same check. If you think the specification is wrong, please report the issue to them. If you have different requirements, consider using a custom method. In case you need to adjust the built-in validation regular expression patterns, please follow the documentation.

PR #1814, opened 2016-07-07
Changes jQuery alias 'expr[":"]' to 'expr.pseudos'

### LINKED ISSUE #1813: Use of $.expr[ ":" ] is deprecated in jQuery 3.0.0
jQuery migration documentation suggests using $.expr.pseudos instead.
https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigrate-jqueryexpr-is-jqueryexprpseudos

kevindb:
Uses newer alias as per jQuery Migrate documentation
https://github.com/jquery/jquery-migrate/blob/master/warnings.md#jqmigra
te-jqueryexpr-is-jqueryexprpseudos
I want to resolve #1813

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -186,7 +186,7 @@ $.extend( $.fn, {
 } );
 
 // Custom selectors
-$.extend( $.expr[ ":" ], {
+$.extend( $.expr.pseudos, {
 
 	// http://jqueryvalidation.org/blank-selector/
 	blank: function( a ) {
--- a/src/core.js
+++ b/src/core.js
@@ -186,7 +186,7 @@ $.extend( $.fn, {
 } );
 
 // Custom selectors
-$.extend( $.expr.pseudos || $.expr[ ":" ], {
+$.extend( $.expr.pseudos || $.expr[ ":" ], {		// '|| $.expr[ ":" ]' here enables backwards compatibility to jQuery 1.7. Can be removed when dropping jQ 1.7.x support
 
 	// http://jqueryvalidation.org/blank-selector/
 	blank: function( a ) {
--- a/src/core.js
+++ b/src/core.js
@@ -186,7 +186,7 @@ $.extend( $.fn, {
 } );
 
 // Custom selectors
-$.extend( $.expr.pseudos, {
+$.extend( $.expr.pseudos || $.expr[ ":" ], {
 
 	// http://jqueryvalidation.org/blank-selector/
 	blank: function( a ) {


staabm:
In which version of jquery was `$.expr.pseudos` added?

staabm:
If I read the commit [1] correctly, this feature `$.expr.pseudos` was added with jquery 3.0.

this means merging this PR would drop support for jquery 2.x ?

[1] https://github.com/jquery/jquery/commit/0402963845be8d71c4e8ddf65e7c055014739b60

Arkni:
@staabm 
AFAIK, `.pseudos` is a `Sizzle` thing and was there from the version `1.8.0` of `Sizzle`.
Also, `$.expr[ ":" ]` is an alias of `$.expr.pseudos`. So it's backward compatible with older versions of `jQuery` (`1.8.0+`).

kevindb:
It looks like the jQuery Foundation does not want to encourage the use of custom pseudos, so public documentation is extraordinarily sparse. As far as I can tell from the discussions linked below, $.expr.pseudos (or jQuery.expr.pseudos) works back to 1.8.0.
   - https://github.com/jquery/jquery/issues/2073
   - https://stackoverflow.com/questions/11624345/getting-the-match-object-in-a-custom-filter-selector-in-jquery-1-8
   - https://github.com/jquery/sizzle/wiki#back-compat
   - https://bugs.jquery.com/ticket/12409

staabm:
Thx for the patch, looks good as is.

Since we support jquery 1.7.2+ we need to wait for a while before merging (or a 1.7.2+ compatible patch).

Arkni:
> ... (or a 1.7.2+ compatible patch)

Something like this ?
```js
$.extend( $.expr.pseudos || $.expr[ ":" ], { ... } );
```

staabm:
Yeah, imagined something like that

kevindb:
Thank you, @Arkni! I added your backwards compatibility fix and did some quick testing. It seems to be working with both jQuery 3.0 and 1.7.2.

Arkni:
My pleasure :)

Just a note, something wrong happened when you rebased your PR (there are 10 unrelated commits included). You can fix that by rebasing a second time on top of `upstream/master`.

kevindb:
*PR is fixed now*
Ah ha! I thought that was weird to see so many commits in the PR, but I wasn't sure how to fix it. Thanks again!

src/core.js
```diff
@@ -186,7 +186,7 @@ $.extend( $.fn, {
 } );
 
 // Custom selectors
-$.extend( $.expr[ ":" ], {
+$.extend( $.expr.pseudos || $.expr[ ":" ], {
```
staabm: Please add a comment that we have a jq 1.7 bc related line here, so we dont miss it when we drop 1.7.x support

staabm:
squashed and merged, thx!

PR #1815, opened 2016-07-07
Select Trigger: trigger onkeyup on select input change

colbyfayock:
Select elements don't really fit the flow of the keyup events on input elements, making it a bit confusing as text inputs directly update with user input, but the selects are not after selection, but until the 2nd action occurs with the focus out.

To provide similar feedback, this triggers the onkeyup method when a "change" event fires only on a select element, passing along the same event and element details to the call.

Previous validation: https://jsfiddle.net/6ymhsae8/1/
1. Hit submit
2. Start typing in text input, form is valid after first keyup
3. Select an option in the select input, user has to focus out of the input to trigger a revalidation

Updated validation: https://jsfiddle.net/x5qk0n9s/1/
1. Hit submit
2. Start typing in text input, form is valid after first keyup
3. Select an option in the select input, upon selecting a valid option, field is no longer marked as invalid

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -311,17 +311,9 @@ $.extend( $.validator, {
 		},
 		onchange: function( element, event ) {
 
-			// If target element is a select, trigger focusout on change
+			// If target element is a select, trigger keyup on change
 			if ( element.tagName.toLowerCase() === "select" ) {
-				var _this = element;
-
-				if ( !_this.form && _this.hasAttribute( "contenteditable" ) ) {
-					_this.form = $( _this ).closest( "form" )[ 0 ];
-				}
-
-				var validator = $.data( _this.form, "validator" ),
-					settings = validator.settings;
-				settings.onkeyup.call( validator, _this, event );
+				this.settings.onkeyup.call( this, element, event );
 			}
 		},
 		highlight: function( element, errorClass, validClass ) {
--- a/src/core.js
+++ b/src/core.js
@@ -309,6 +309,21 @@ $.extend( $.validator, {
 				this.element( element.parentNode );
 			}
 		},
+		onchange: function( element, event ) {
+
+			// If target element is a select, trigger focusout on change
+			if ( element.tagName.toLowerCase() === "select" ) {
+				var _this = element;
+
+				if ( !_this.form && _this.hasAttribute( "contenteditable" ) ) {
+					_this.form = $( _this ).closest( "form" )[ 0 ];
+				}
+
+				var validator = $.data( _this.form, "validator" ),
+					settings = validator.settings;
+				settings.onkeyup.call( validator, _this, event );
+			}
+		},
 		highlight: function( element, errorClass, validClass ) {
 			if ( element.type === "radio" ) {
 				this.findByName( element.name ).addClass( errorClass ).removeClass( validClass );
@@ -395,7 +410,7 @@ $.extend( $.validator, {
 			}
 
 			$( this.currentForm )
-				.on( "focusin.validate focusout.validate keyup.validate",
+				.on( "focusin.validate focusout.validate change.validate keyup.validate",
 					":text, [type='password'], [type='file'], select, textarea, [type='number'], [type='search'], " +
 					"[type='tel'], [type='url'], [type='email'], [type='datetime'], [type='date'], [type='month'], " +
 					"[type='week'], [type='time'], [type='datetime-local'], [type='range'], [type='color'], " +


src/core.js
```diff
@@ -309,6 +309,21 @@ $.extend( $.validator, {
...
+			if ( element.tagName.toLowerCase() === "select" ) {
+				var _this = element;
+
+				if ( !_this.form && _this.hasAttribute( "contenteditable" ) ) {
+					_this.form = $( _this ).closest( "form" )[ 0 ];
+				}
+
+				var validator = $.data( _this.form, "validator" ),
+					settings = validator.settings;
+				settings.onkeyup.call( validator, _this, event );
```
Arkni: `316 -> 324` can be shortened to:
```js
this.settings.onkeyup.call( this, element, event );
```

Also, why a `select` element should be **editable** ?

Arkni:
By the way, the jsfiddle links are the same

src/core.js
colbyfayock: Sure, I'll look into working that in tomorrow, though if I remember correctly, I did it that way because the context wasn't right.

Are you referring to ```_this.hasAttribute( "contenteditable" )```? If so, I was mostly moving the same logic from the original event into this block so I didn't miss anything.

colbyfayock:
Ah, I blame my clipboard! :) I fixed it. New fiddle is https://jsfiddle.net/x5qk0n9s/1/

src/core.js
Arkni: > Are you referring to _this.hasAttribute( "contenteditable" )?

Yes


> I was mostly moving the same logic from the original event into this block so I didn't miss anything.

No need for it.
Just noticed that that block is redundant with the one in function `delegate`, so you can omit it.

colbyfayock:
Worked perfect, updated. Thanks

src/core.js
```diff
@@ -309,6 +309,13 @@ $.extend( $.validator, {
 				this.element( element.parentNode );
 			}
 		},
+		onchange: function( element, event ) {
```
staabm: I am not sure this fix will work in all situations. 

this code will change only the "default-onchange" implementation. in case someone has overridden this option he will not benefit from your change.

colbyfayock: Wouldn't that be the same for any of the event handlers though? Being able to override and lose the default behavior unless set up in the override?

If that's an issue, another option, I can throw the logic into delegate, but originally I avoided that to prevent throwing in specific event logic into that function.

Or any other ideas?

staabm: I am not sure yet, whats the correct approach ;-).

staabm:
just to make sure I get what the reported problem is:

the bug only happens when using the keyboard with TAB to jump into the select and use the arrow keys to select a option, right?

using the mouse to select a option works as expected in the provided "previous" example?

colbyfayock:
No, it's with mouse click. It appears it works properly in Firefox, but it doesn't in Chrome, at least with my testing and a coworker's for a sanity check.

staabm:
In my Chrome52 I do the following:
- open https://jsfiddle.net/6ymhsae8/1/ in incognito
- click submit
- start typing into input field (focus was autom. set into it)
- click to open the select-box
- select "One"
--> the validation message immediately disappears

(same on FF49)

colbyfayock:
:( weird. here's gifs of it not working for my coworker and i

- Chrome Version 52.0.2743.75 beta (64-bit) => http://d.fay.io/M4Fy
- Chrome Version 50.0.2661.102 (64-bit) => http://d.pr/i/ILUG
- He also just had an update and tested it again on Version 51.0.2704.106 (64-bit) with it still not working

And just to confirm, it's working for me in FF

 - Firefox 43.0.4 => http://d.fay.io/QH6u

staabm:
looks like you are testing on mac and I tested it on win7x64

@Arkni are you able to reproduce the issue?

Arkni:
> the bug only happens when using the keyboard with TAB to jump into the select and use the arrow keys to select a option, right?

Yes, only appears in this case. Otherwise, everything works as expected.

> @Arkni are you able to reproduce the issue?

No. I got the same results as you. I testes on `Chrome 50.0.2661.102`, `Firefox 64.0.1` and `Firefox 49` on `Ubuntu 14.04 LTS`.

colbyfayock:
I have two Macs specifically confirmed, are there any other contributors on a mac?

Arkni:
Googled a little and found some open bugs on Chromium bug tracker related to the `click` event. These 2 bug reports prove that the `click` event on `SELECT` elements (or on `OPTION` on all machines) is not working properly on `Chrome` on Mac machines:

* https://bugs.chromium.org/p/chromium/issues/detail?id=5284
* https://bugs.chromium.org/p/chromium/issues/detail?id=26017

I think a fair solution/workaround will be to bind the `change` event to `SELECT` elements only. This way, we will make sure that only the `SELECT` elements will be validated on `change` event.

staabm:
Maybe we can do a feature detect and patch the elements only in browers not working correctly right now?

colbyfayock:
So while the current pull binds the change event to all, it only actually triggers validation, or really does anything, on the select elements, which just forwards it to the keyup handler, but it does add the extra handler on everything, which isn't ideal.

An argument to keep it as is, if I'm not mistaken, leaving the onchange method now opens it up for developers to hook into which is a plus, in my opinion, unless it's intentionally left out.

All that said, I'd be happy to pull it out and bind the base event to select. I can look into the feature detection side of it as well.

staabm:
Since noone was missing the possibility of a onchange handler I would not add another feature without knowing a good use-case for it.

I would welcome a feature detected workaround.

staabm:
any chanche to get a working unit test for this patch?

colbyfayock:
Sure, I'll try to work those in.

rfkarlin:
Are there any plans to add this change?  I am running into this problem - I want to clear validation on a select when the user selects a good value (before they tab out) and version 1.17.0 of jquery-validation is not doing this for me on mac in either firefox or chrome.  This fiddle does work for me though: https://jsfiddle.net/x5qk0n9s/1/.

colbyfayock:
happy to resurface on this if it's a welcome addition from the maintainers. unless the code itself needs changing, i think im mostly just missing the unit test

rfkarlin:
That would be great.  I think it's going to confuse our users when they select a valid option and the error message doesn't immediately go away.  So far I haven't found any work around.

Issue #1816, opened 2016-07-13
Wrong validation in groups where 3 and more fields


vkurlyan:
### Wrong validation in groups where 3 and more fields
If group contain 3 and more fields, we try to submit invalid form, then trigger blur event on first selectbox. Result: third selectbox has valid class, but it has invalid value.

### Your environment
* jquery-validate 1.15.0
* chrome 51.0.2704.106 (64-bit)

### Steps to reproduce
* open https://jsfiddle.net/vgtba9u1/9/
* click submit, all selectoboxes have error class.
* click on any empty place to trigger blur event on "day" selectbox.

### Expected behaviour
all selectboxes still have error class

### Actual behaviourear
last selectbox has 'valid' class

Note: I am going to make pull request

staabm:
@vkurlyan please verify the landed patch fixes your issue

vkurlyan:
yes, it does

PR #1805, opened 2016-06-24
Fix group validation

TLib:

PR #1817, opened 2016-07-13
Fix validation of group where 3 and more fields

vkurlyan:
fix issue https://github.com/jzaefferer/jquery-validation/issues/1816


Arkni:
Duplicate of #1805

PR #1823, opened 2016-07-21
Select2 support

waitingsong:
errorElement append after span.select2-container if <select> element init by jquery.select2

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -925,7 +925,17 @@ $.extend( $.validator, {
 				} else if ( this.settings.errorPlacement ) {
 					this.settings.errorPlacement.call( this, place, $( element ) );
 				} else {
-					place.insertAfter( element );
+					var $elm = $(element),
+						$h;
+
+					if ($elm.hasClass('select2-hidden-accessible')) {	// hidden by jquery.select2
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -925,7 +925,17 @@ $.extend( $.validator, {
...
+						$h;
+
+					if ($elm.hasClass('select2-hidden-accessible')) {	// hidden by jquery.select2
+						$h = $elm.next('.select2-container');
+						$h[0] && place.insertAfter($h);
+					}
+					else {
+						place.insertAfter(element);
+					}
+					$elm = $h = null;
```
Arkni: Why not using [`errorPlacement`](https://jqueryvalidation.org/validate/#errorplacement) option to do that ?
```js
$("#form").validate({
  rules: { ... },
  messages: { ... },
  errorPlacement: function ( error, element ) {
    // Your code goes here
  }
});
```

staabm:
we wont land code which is specific to the markup of any jsframework/lib out there. you should work with the provided options/settings jquery-validate provides (like Arkni mentioned above).

I will close this issue for now, as the patch is not acceptable.
Feel free to ask further questions for your particular problem though.

PR #1843, opened 2016-08-18
Always bind invalidHandler, do so anonymously

YodaDaCoda:
This pr allows changing of invalidHandler after initialisation, and setting of invalidHandler if initialised without.

Fiddle:
https://jsfiddle.net/YodaDaCoda/fonmoyck/
Note that it's the initial invalidHandler that fires, not the replacement.

Fiddle with this patch:
https://jsfiddle.net/YodaDaCoda/Lugjrgge/
Note that the replacement invalidHandler fires.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -405,9 +405,11 @@ $.extend( $.validator, {
 				// "select" is provided as event.target when clicking a option
 				.on( "click.validate", "select, option, [type='radio'], [type='checkbox']", delegate );
 
-			if ( this.settings.invalidHandler ) {
-				$( this.currentForm ).on( "invalid-form.validate", this.settings.invalidHandler );
-			}
+			$( this.currentForm ).on( "invalid-form.validate", function (event, validator) {
+				if ( validator.settings.invalidHandler ) {
+					validator.settings.invalidHandler( event, validator );
+				}
+			});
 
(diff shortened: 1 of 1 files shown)


src/additional/phoneAU.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Australian phone numbers have 10 digits (or 11 and start with +o31).
+ */
+$.validator.addMethod( "phoneNL", function( value, element ) {
```
staabm: Doesnt belong into thos PR?

staabm:
Whats your use case that requires this change?

src/additional/phoneAU.js
YodaDaCoda: Sorry, I'm pretty bad at git sometimes.

YodaDaCoda:
I use the validator as part of a package which is deployed to clients with some default settings (the deployed code can't be changed, only added to). When clients want something that requires a modified invalidHandler, I encounter difficulty as I'm not able to change or remove the default action.

staabm:
Sorry, but I dont think this change will progress the library into a good direction.

staabm:
Will close for now. Feel free to ask further questions regarding this topic here. We can re-open if required.

PR #1850, opened 2016-08-31
Polish tax id validation method

krzywiecki:
This PR adding new custom method to validate Polish tax identifier number (NIP).

Working example:
https://jsfiddle.net/dycqm91k/1/


staabm:
Thx

Issue #1806, opened 2016-06-26
[Request] promise support


happyDemon:
### Subject of the issue

I'm writing an image dimension validation rule (for which I have to wait for the image to load in order to be able to check its width & height). Normally I'd make a promise and return that while the image is being loaded. Once the image is loaded it would be easy to resolve or reject the promise based on the defined dimension restrictions.

I thought the remote rule would be the same (since out of the box it returns a promise). but it only seems to return 'pending'.

is there an easy way for me actually implement this rule? or is a promise-based return value on the schedule?

I have implemented this in parsleyJS which was able to take a promise as a return, so I'm kind of confused why it's not possible here.
### Your environment

n/a
### Steps to reproduce

n/a
### Expected behaviour

n/a
### Actual behaviour

n/a

staabm:
This library pre-dates promises in JS, so thats the reason why those are not yet supported.

I am open to discuss how a promise integration could look like though.

happyDemon:
hopefully it shouldn't be too hard, switching out the "pending" handler with a promise handler is basically what needs to be done.
This promise handler would need to support jquery 1.x & 2.x flavoured promises and the standard JS promises (also featured in jquery 3).
After that a little rewrite of the remote rule and we're done.

I can get cracking on a pull request.

staabm:
I would be ok to only support ES6 promises.

The change needs to be backwards compatible though.

santiagodoldan:
@happyDemon were you able to implement something? I'd like to help if possible

Issue #1763, opened 2016-04-09
JQuery Validation, call the valid method, is not call remote


lcyan:
JQuery Validation, call the valid method, is not call remote, remote valid is not work.

$id_mobile.valid()

Arkni:
Please provide a [JSFiddle](https://jsfiddle.net/) or [JSBin](https://jsbin.com/) containing a set of reproducible steps.

toabi:
I think this might be a duplicate of https://github.com/jzaefferer/jquery-validation/issues/1045

staabm:
Agree, duplicate of #1045.

Will close for now. Feel free to ask further questions regarding this topic here. We can re-open if required.

Issue #1465, opened 2015-04-20
valid() and element() incorrectly return true for remote method


NotionCommotion:
The first time `valid()` and `element()` is executed (for example, in the click event on testit), they both incorrectly return true even if the remote method indicates that it shouldn't pass validation.  Note that the error message in fact is still displayed.

    <!DOCTYPE html>
    <html>
        <head>
            <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
            <title>Testing</title>  
            <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.2/jquery.js" type="text/javascript"></script>
            <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-validate/1.13.1/jquery.validate.js" type="text/javascript"></script>
            <script type="text/javascript"> 
                $(function(){
                    var validator=$("#myForm").validate({
                        rules: {
                            bla: {
                                minlength:2,
                                maxlength:4,
                                required:true,
                                remote: "validate.php"
                            }
                        },
                        messages: {
                            bla: {
                                remote:"error message"
                            }
                        }
                    });
    
                    $('#testit').click(function(){
                        $('#bla').val('test');
                        console.log($('#bla').valid());
                        console.log(validator.element('#bla'));
                    })
                });
            </script>
        </head>
    
        <body>
            <form id="myForm" method="post">
                <input name="bla" id="bla" value="" readonly>
            </form>
            <button id="testit">testit</button>
        </body> 
    </html>

NotionCommotion:
Looks like this has already been identified at https://github.com/jzaefferer/jquery-validation/issues/1045.  Any idea whether it will ever be addressed?

jzaefferer:
Closing in favor of #1045

Issue #1344, opened 2014-12-06
Is there an async version of .valid() method?


yulanggong:
When deal with multiple remote validations, I hope the `valid` method return a promise object.

like this.

``` js
$form.valid()
.done(function(){
  //
})
.fail(function(){
  //
})
```

PR #1854, opened 2016-09-04
Add warning message support for pending remote validations

### LINKED ISSUE #823: remote return false, but form.valid() return true(expect return false)
when the form field has value and the remote ajax(jsonp) request return false, invoke the $form.valid() return true(expect return false)

here is the example  http://jsbin.com/eresut/22/watch

aek:
Component: Core
Add warning message support for pending remote validations while keeping the field and form as invalid
Fixes #823 #1044 #1045 #1283 #1344 #1465 #1763 #1806

This tweaks will allow you to show a warning message to the user while the async response to the remote validations get resolved and it will make the valid response to be false if the async response to the remote validation is not ready yet but also it will allow to auto clear the invalid state for the field and the form once the async response is ready by using the remote validation normal flow so if the response is valid the field value is valid and the warning message get cleared automatically by jquery validation or if the response is not valid the error message defined is used instead of the warning message.
Just need to add another rule entry for the custom warning messages or the error default messages will be used instead. 
I think that this is the most simply way to solve this problem without having to rewrite the plugin to make it work with deferred/promises. It's a very desirable feature to have deferred/promises support but this will solve the issues and also it could be used for dependency mismatch validations too

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -348,6 +348,24 @@ $.extend( $.validator, {
 		min: $.validator.format( "Please enter a value greater than or equal to {0}." ),
 		step: $.validator.format( "Please enter a multiple of {0}." )
 	},
+	warnings: {
+		required: "This field is required.",
+		remote: "Please fix this field.",
+		email: "Please enter a valid email address.",
+		url: "Please enter a valid URL.",
+		date: "Please enter a valid date.",
+		dateISO: "Please enter a valid date (ISO).",
+		number: "Please enter a valid number.",
+		digits: "Please enter only digits.",
+		equalTo: "Please enter the same value again.",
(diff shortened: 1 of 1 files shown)


aek:
Please review the errors in travis-ci checks since they are referring to stuffs available in the core itself like comments already available

staabm:
Thanks for your work.

I dont agree that we should add the concept of warnings which more or less duplicates most stuff our error handling is already capable of.

The only real solution for the "remote pending problem" would be to use a promise based implementation. Thats something we could only add in a new major version because it will not be backwards compatible.

staabm:
Will close for now because of inactivity. Feel free to ask further questions/provide furhter information regarding this topic here. We can re-open if required.

PR #1858, opened 2016-09-08
Move jquery dependencies to dev dependencies

svarnoi420:
I installed jquery-validation from `npm`. After compiled with `webpack` I have two version jquery on my bundle, first - mine, second - it is the plugin jquery dependencies. And so I moved  this dependencies to dev dependencies.

Weird, why this package have own `node_modules` on folder after installed with `npm`

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,6 @@
 		"dist/jquery.validate.js"
 	],
 	"main": "dist/jquery.validate.js",
-	"dependencies": {
-		"jquery": "^1.7 || ^2.0"
-	},
 	"devDependencies": {
 		"commitplease": "2.3.1",
 		"grunt": "1.0.1",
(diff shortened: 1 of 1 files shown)


staabm:
Jqery is not a dev but a runtime dependency for jquery-validate.

How did you declare your other jquery dependency?

svarnoi420:
@staabm I installed them from `npm` and set aliases `$`, `jquery`, `jQuery` for requiring like this `require('jquery')` and etc.

As `jquery-validation` have own node_modules folder, it defining jquery from there.

staabm:
the version which is available via npm is managed in this repo.

tswaters:
If it's not a direct dependency it should probably be a `peerDependency`  -- it is expected to be present, but it is up to the user to have whatever jquery as their own dependency.

Guymestef:
@svarnoi420 What's the jquery version you're using in your package.json?

I have this problem too, and I think it's because I use the latest jquery version (v3) but jquery.validation has a dependency on jquery ^1.7 || ^2.0. So webpack build jquery.validation with jquery v2 and all other libs with jquery v3.

svarnoi420:
@Guymestef Yeah, I use jquery v3 too. Hm, I didn't think about this can be reason. I'll try install jquery v2 for analyze.

staabm:
I just landed a PR which added support for jquery3.x, will close therefore. feel free to provide further information in case this was not resolved by this change.

PR #1868, opened 2016-09-23
fix parameter

AbdullahObaid:
the message is not appearing probably because {0} is written (0}


staabm:
merged via https://github.com/jzaefferer/jquery-validation/commit/c405c62b3cf39bf22cbce045a099964cada91280

thx for the heads up.

PR #1892, opened 2016-11-15
Init: add binding for input with button type

### LINKED ISSUE #1891: Input with type=button validated but not focused in
### Subject of the issue
Hi guys. We've been using validator in scope of our project and discovered one interesting think. 
If input type is button then it can be validated by validation rules, but handler on `focusin` event will not be fired. It happens because of 
https://github.com/jzaefferer/jquery-validation/blob/master/src/core.js#L398-L402 
handler for `type=button` is not set.

### Environment
* version of `jquery-validate` - 1.15.1
* `Chrome Canary` - browser

### Steps to reproduce
Here some fiddle - https://jsfiddle.net/yvmd3nta/

I have several ideas how to solve this.

1. Add handler for `type=button`
2. Add option where list of elements which will be handled can be set. Default will be current list.
3. Remove validation from inputs with `type=button`

I would prefer either 1 or 2 and like to make PR if you don't mind.

denar90:
Resolves #1891 

I'm adding the easiest solution from list described in issue.
Here example with build based on my changes - https://jsfiddle.net/5vumyq0k/

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -399,7 +399,7 @@ $.extend( $.validator, {
 					":text, [type='password'], [type='file'], select, textarea, [type='number'], [type='search'], " +
 					"[type='tel'], [type='url'], [type='email'], [type='datetime'], [type='date'], [type='month'], " +
 					"[type='week'], [type='time'], [type='datetime-local'], [type='range'], [type='color'], " +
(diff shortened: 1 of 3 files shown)


staabm:
Thx

PR #1898, opened 2016-11-19
Core: Added callback for access element and modify data in remote rule.

christianrss:
Added callback for access the element dinamically and modify data in remote rule.

Example:
My need was remove mask of value before send data to server.

It could be done with:
```
remote: {
    ...
    data: function() {
        
    }
    ...
}
```

but I need to access the elements dynamically like this:

```
remote: {
    ...
    parseData: function( element ) {
        data = {};
        data[ element.name ] = $( element ).inputmask("unmaskedvalue");

        return JSON.stringify( data );
    }
    ...
}
```

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1479,8 +1479,15 @@ $.extend( $.validator, {
 			previous.old = optionDataString;
 			validator = this;
 			this.startRequest( element );
-			data = {};
-			data[ element.name ] = value;
+
+			// Identify data callback
+			if ( typeof param.parseData === "function" ) {
+				data = param.parseData( element );
+			} else {
+				data = {};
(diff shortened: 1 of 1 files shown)


Arkni:
No need to modify the `remote` method, you can achieve the same by using the [`normalization layer`](https://jqueryvalidation.org/normalizer/) which only available on versions >= `1.15.0`:
```js
remote: {
  normalizer: function(value) {
    // `this` inside the normalizer is the element instance
    // `value` is the current value of the element (you wont need it in your case)

    // The returned value will be used to validate the current element
    return $(this).inputmask("unmaskedvalue");
  }
}
```

christianrss:
Yes, you right. Thanks.

PR #1901, opened 2016-12-03
Added support for button elements with descendants

phistuck:
JSFiddle -
https://jsfiddle.net/0b3we0gm/9/ (with the fix)
https://jsfiddle.net/0c8nvnob/7/ (without the fix)

Non-Firefox browsers dispatch `click` on descendants of `<button>`, which breaks
the submitHandler logic that adds a hidden field with the name and value of the
submit button (since it is not included in the submission if it did not
initiate the submission, which might be the case when using `form.submit` in
`submitHandler` and returning `false`). `submitButton` is instead a descendant of
`<button>` and the hidden field does not compensate for the lack of an actual
submit button.

This would fix https://crbug.com/668524 and make `submitButton` an actual button.

The issue is reproducible in Chrome.
Firefox does not dispatch events to the descendants of `<button>` (but the specification requires that it will, see https://bugzilla.mozilla.org/show_bug.cgi?id=1089326 for details). The rest of the gang dispatch events to the descendants, so I imagine the problem exists in all of them and not just in Chrome.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -397,6 +397,28 @@ QUnit.test( "submitHandler keeps submitting button", function( assert ) {
 	$( "#userForm" ).submit();
 } );
 
+QUnit.test( "submitHandler keeps submitting button, even if descendants are clicked", function( assert ) {
+	var button, event, buttonDescendant;
+
+	$( "#testForm27" ).validate( {
+		debug: true,
+		submitHandler: function( form ) {
+
+			// Dunno how to test this better; this tests the implementation that uses a hidden input
+			var hidden = $( form ).find( "input:hidden" )[ 0 ];
+			assert.deepEqual( hidden.value, button.value );
+			assert.deepEqual( hidden.name, button.name );
+		}
+	} );
+	$( "#testForm27 [name=\"year\"]" ).val( "2016" );
+	button = $( "#testForm27 :submit" )[ 0 ];
+	buttonDescendant = $(button).find("span");
+	event = $.Event( "click" );
+	event.preventDefault();
+	$.event.trigger( event, null, buttonDescendant );
+	$( "#testForm27" ).submit();
+} );
+
 QUnit.test( "validation triggered on radio/checkbox when using keyboard", function( assert ) {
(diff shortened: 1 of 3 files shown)


phistuck:
An alternative approach is to use `currentTarget` instead of `target`. Thoughts?

phistuck:
The `currentTarget` approach seems more reasonable and simpler, so I went ahead with it.

phistuck:
@davbizz - I did already. See https://github.com/jzaefferer/jquery-validation/pull/1901/commits/c257fc0de1a093af8c4bae9e04951291d6482ce1.

test/test.js
```diff
@@ -397,6 +397,26 @@ QUnit.test( "submitHandler keeps submitting button", function( assert ) {
...
+			assert.deepEqual( hidden.name, button.name );
+
+			return false;
+		}
+	} );
+	$( "#testForm27 [name=\"year\"]" ).val( "2016" );
+	button = $( "#testForm27 :submit" )[ 0 ];
+	$( button ).find( "span" ).click();
+} );
+
```
Arkni: You can extend your tests by testing the property `submitButton` of the validator. So, your tests will look like this:
```js
var button = $( "#testForm27 :submit" )[ 0 ];
var v = $( "#testForm27" ).validate( {
	debug: true,
	submitHandler: function( form ) {

		// Compare the button with the `submitButton` property
		assert.deepEqual(
			v.submitButton, button, "The submitButton property should be the same as button"
		);

		var hidden = $( form ).find( "input:hidden" )[ 0 ];
		assert.deepEqual( hidden.value, button.value );
		assert.deepEqual( hidden.name, button.name );

		return false;
	}
} );

$( "#testForm27 [name=\"year\"]" ).val( "2016" );

$( button ).find( "span" ).click();
```

phistuck: I added your test as well.

staabm:
Thx for your fix.

I will merge this now but have mixed feelings about it. Be warned that we could revert this again in case it breaks something...

PR #1905, opened 2016-12-11
Core: Add support for defining a global normalizer

Arkni:
This PR added support for using a global normalizer for all elements of the form, which will run for all elements if it's present in the top level (the same level as `rules`) without affecting the current behaviour, that way you can transform the values of a set of elements without attaching the same normalizer to the aforementioned elements, and you can override it on element level by specifying a normalizer for the specific element that you want to transform differently.

In pseudo code, this will look like:

``` js
$("#myForm").validate({
  normalizer: function(value) {
    // Trim the value of every element
    return $.trim(value);
  },
  rules: {
    username: {
      required: true,
      // Here, we override the global normalizer
      normalizer: function(value) {
        value = value.replace("something", "new thing");
        return $.trim(value);
      }
    }
  }
});
```

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -717,21 +717,27 @@ $.extend( $.validator, {
 				} ).length,
 				dependencyMismatch = false,
 				val = this.elementValue( element ),
-				result, method, rule;
+				result, method, rule, normalizer;
 
-			// If a normalizer is defined for this element, then
-			// call it to retreive the changed value instead
+			// Prioritize the local normalizer defined for this element over the global one
+			// if the former exists, otherwise user the global one in case it exists.
+			if ( typeof rules.normalizer === "function" ) {
+				normalizer = rules.normalizer;
+			} else if (	typeof this.settings.normalizer === "function" ) {
+				normalizer = this.settings.normalizer;
+			}
+
+			// If normalizer is defined, then call it to retreive the changed value instead
 			// of using the real one.
 			// Note that `this` in the normalizer is `element`.
-			if ( typeof rules.normalizer === "function" ) {
-				val = rules.normalizer.call( element, val );
+			if ( normalizer ) {
+				val = normalizer.call( element, val );
 
 				if ( typeof val !== "string" ) {
 					throw new TypeError( "The normalizer should return a string value." );
 				}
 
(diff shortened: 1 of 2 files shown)


staabm:
Thx.🎄

cheonhyangzhang:
Hi @Arkni 
when you have defined global normalizer and specific normalizer, will they both be executed? or only one of them is executed?    If both is executed, what's the order of execution?

Arkni:
Only one of them is executed which is the specific one.
What we do here is overriding the global normalizer in case a specific one exist. So in the following example, only `usernameNormalizer` will be executed for the `username` element and  `globalNormalizer` will run for other fields:
```js
$("#myForm").validate({
  // This will run for every element except for the ones that have their own normalizer.
  // In our case, this will not run for `username` element.
  normalizer: function globalNormalizer(value) {
    return $.trim(value);
  },
  rules: {
    username: {
      required: true,
      // Here, we override the global normalizer
      normalizer: function usernameNormalizer(value) {
        return value.toUpperCase();
      }
    },
    fullname: {
      required: true
    }
  }
});
```

cheonhyangzhang:
@Arkni   Got it.

Thanks for the explanation and the nice work.

Do you know about any schedule for the next release? Is next release going to have this feature? And if so when will that be ?

Arkni:
@cheonhyangzhang 

> Do you know about any schedule for the next release?

There still some issues to be fixed before the next release. So, maybe after fixing them, we will do a minor release.

> Is next release going to have this feature? 

Yes

> And if so when will that be ?

Same as answer 1.

cheonhyangzhang:
@Arkni 
Cool!   Thanks!

reachmehar:
Nice one. Any ETA on this feature? 
According to https://jqueryvalidation.org/normalizer/ it seems like both global and element specific normalizer might work but not yet released?

Arkni:
The new release was planned for the end of June but got postponed and the documentation is ahead of its time :)

We will see if we can prepare a release the next weekend!

/c @staabm

reachmehar:
That's nice.. how to subscribe for release alerts ?

Arkni:
Subscribe to #2016

PR #1932, opened 2017-01-30
README: removed gitter badge

staabm:
I dont use it.. anyone else?


Arkni:
I don't use it myself, but if people want help, they can ask for it here or on StackOverflow.

staabm:
Agree, we shouldnt promote a channel in which no core contributor is present.

PR #1936, opened 2017-02-01
Added Step Range Validation  to messages_ja

Worthy7:
Added Step Validation message


src/localization/messages_ja.js
```diff
@@ -18,6 +18,7 @@ $.extend( $.validator.messages, {
 	minlength: $.validator.format( "{0} 文字以上で入力してください。" ),
 	rangelength: $.validator.format( "{0} 文字から {1} 文字までの値を入力してください。" ),
 	range: $.validator.format( "{0} から {1} までの値を入力してください。" ),
+	step: $.validator.format("{0} の倍数を入力してください。"),
```
Arkni: Can you add one space after `(` and one before `)` as done for other lines. So the line should look like:
```js
step: $.validator.format( "{0} の倍数を入力してください。" ),
```

staabm:
Thx

PR #1942, opened 2017-02-04
Core: count invalid fields with empty message in `numberOfInvalids()`

### LINKED ISSUE #1922: numberOfInvalids() does not count invalid fields that have no error message
Don't know if this behavior is intended or not, but I think it's at least questionable, hence my issue report.

So, method `numberOfInvalids()` does not count invalid fields _that have no error message_. I think it's perfectly valid to have empty error message for certain type of validation, depending on the way errors are indicated; and an invalid field with zero-length error message is still invalid. However,  method `objectLength()` called by `numberOfInvalids()` only counts non-empty values:

```javascript
objectLength: function( obj ) {
	var count = 0, i;
	for ( i in obj ) {
		if ( obj[ i ] ) {
			count++;
		}
	}
	return count;
}
```

why is it so? Shouldn't the condition `if ( obj[ i ] )` be replaced with, say, `if ( typeof obj[ i ] != 'undefined')` or just dropped?

Arkni:
Fixes #1922

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -552,7 +552,10 @@ $.extend( $.validator, {
 			var count = 0,
 				i;
 			for ( i in obj ) {
-				if ( obj[ i ] ) {
+
+				// This check allows counting elements with empty error
+				// message as invalid elements
+				if ( obj[ i ] !== undefined && obj[ i ] !== null && obj[ i ] !== false ) {
 					count++;
 				}
 			}
--- a/test/test.js
+++ b/test/test.js
@@ -313,6 +313,31 @@ QUnit.test( "check(): simple", function( assert ) {
 	assert.ok( v.size() === 0, "No more errors" );
 } );
 
+QUnit.test( "numberOfInvalids(): count invalid fields with empty message", function( assert ) {
+	assert.expect( 2 );
+
+	var v = $( "#testForm23" ).validate( {
+		rules: {
+			box1: {
+				required: true
+			},
+			box2: {
+				required: true
+			}
+		}
+	} );
+
+	var requiredMsg = $.validator.messages.required;
+	$.validator.messages.required = "";
+
+	v.form();
+
+	assert.equal( v.errorList.length, 2, "The form has two errors" );
+	assert.equal( v.numberOfInvalids(), 2, "The form has two errors" );
+
+	$.validator.messages.required = requiredMsg;
+} );
+
 QUnit.test( "hide(): input", function( assert ) {
 	assert.expect( 3 );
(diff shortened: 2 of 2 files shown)


test/test.js
```diff
@@ -313,6 +313,32 @@ QUnit.test( "check(): simple", function( assert ) {
...
+	} );
+
+	var requiredMsg = $.validator.messages.required;
+	$.validator.messages.required = "";
+
+	// Validate the form manually
+	v.form();
+
+	assert.equal( v.errorList.length, 2, "There is only one error" );
+	assert.equal( v.numberOfInvalids(), 2, "There is only one error" );
```
staabm: The message and the expected equals of 2 is missleading

test/test.js
```diff
@@ -313,6 +313,32 @@ QUnit.test( "check(): simple", function( assert ) {
...
+
+	$.validator.messages.required = requiredMsg;
```
staabm: Isnt the validator reset by the testrunner automatically back to defaults after ea h test?

test/test.js
Arkni: Ops my bad, I forgot to update the message because I used to have only one error before adding another element to the list. I'll update it soon

test/test.js
Arkni: Nope. In this case we operate on a global variable which the test runner can't afford to update or reset.

Arkni:
Updated the PR!

test/test.js
```diff
@@ -313,6 +313,32 @@ QUnit.test( "check(): simple", function( assert ) {
...
+			box2: {
+				required: true
+			}
+		}
+	} );
+
+	var requiredMsg = $.validator.messages.required;
+	$.validator.messages.required = "";
+
+	// Validate the form manually
```
staabm: Useless comment

src/core.js
```diff
@@ -552,7 +552,7 @@ $.extend( $.validator, {
 			var count = 0,
 				i;
 			for ( i in obj ) {
-				if ( obj[ i ] ) {
+				if ( obj[ i ] !== undefined && obj[ i ] !== null && obj[ i ] !== false ) {
```
staabm: Please add a comment that we do this verbose check for the empty string exception

test/test.js
Arkni: This is was a leftover of copy-paste from an old test.
Removed it.

Arkni:
Done!

staabm:
Well done, thx @Arkni

PR #1956, opened 2017-02-21
Localization: Ignore message dot

stadniklksndr:
Ignore message dot sign in the end of sentence

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -264,6 +264,7 @@ $.extend( $.validator, {
 		onsubmit: true,
 		ignore: ":hidden",
 		ignoreTitle: false,
+    ignoreMsgDot: false,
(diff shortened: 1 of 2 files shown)


staabm:
why do we need this change and which problem does it solve?

stadniklksndr:
In some projects we need to ignore default errors messages dot sign in the end of sentence.
Why we should do it manually if we can do it by plugin.
This pull gives us the opportunity to solve it without adding custom messages and keep code clean.

staabm:
atm your patch chops the last character, no matter whether its a dot or not.

I also dont agree that this is something which should be fixed with an additional option ( we already have too many options).

in case you need custom message you should provide your own tranlsations/message-bundle IMO.

Will close for now. We can re-open if required.

PR #1966, opened 2017-03-04
Additional: add optional support to cifES, nifES and nieES

cbrianso:
The cifES, nifES and nieES not check if field is optional, and if field is empty show a error message.


Arkni:
And sorry for taking so long to merge this!

PR #1973, opened 2017-03-09
isHexColor.js

miladganjali:
In this additional file we check is input value is hex formate color

### DIFF
--- a/src/additional/ishexcolor.js
+++ b/src/additional/ishexcolor.js
@@ -1,0 +1,3 @@
+$.validator.addMethod( "ishexcolor", function( value, element ) {
+	return this.optional( element ) || /^#(?:[0-9a-f]{6}|[0-9a-f]{3})$/i.test( value );
+}, "This is not hex format color" );


src/additional/ishexcolor.js
```diff
@@ -0,0 +1,3 @@
+$.validator.addMethod( "ishexcolor", function( value, element ) {
```
Arkni: Can you change the name to `hexcolor` in order to be consistent with other additional methods?
Also, I would like the file to be renamed to `hexcolor` as well.

src/additional/ishexcolor.js
```diff
@@ -0,0 +1,3 @@
...
+	return this.optional( element ) || /^#(?:[0-9a-f]{6}|[0-9a-f]{3})$/i.test( value );
+}, "This is not hex format color" );
```
Arkni: Please change this message to something like: `Please specify a valid hex color value`

PR #1985, opened 2017-03-31
Rip747 rasied errors

### LINKED ISSUE #1925: Get invalid rule type
Is there a callback that can be used to get the specific rule that is invalid? For example, I have an input with two range rules. The first one is a "warning" [-5,5] where an input box is highlighted in yellow. The second one is "danger" [-10, 10] where an input box is highlighted in red.

In `.validate()` I cannot seem to find a callback that will allow me to determine which specific rule key is invalid. I really want to avoid using `errorPlacement` to search the error message in order to determine the specific rule.

```
$(document).ready(function() {

  $.validator.addMethod(
    'rangeWarning',
    function(value, element, parameters) {
      var valid = false;
      if (value >= parameters[1] && value <= parameters[2]) {
        valid = true;
      }
      return this.optional(element) || valid;
    },
    $.validator.format("{0} - value should be between {1} and {2}")
  );
  
  $.validator.addMethod(
    'rangeDanger',
    function(value, element, parameters) {
      var valid = false;
      if (value >= parameters[1] && value <= parameters[2]) {
        valid = true;
      }
      return this.optional(element) || valid;
    },
    $.validator.format("{0} - value should be between {1} and {2}")
  );  

  validator = $('#myform').validate({
    rules: {
      field1: {
        required: true,
        rangeDanger: ['danger', -10, 10],
        rangeWarning: ['warning', -5, 5]
      }
    },
    unhighlight: function(element, errorClass, validClass) {
    	$(element).removeClass('danger').removeClass('error').removeClass('warning').addClass(validClass);
    },   
    errorPlacement: function (error, element) {
    	$(element).removeClass('danger').removeClass('warning');
    	if (error.text().indexOf('warning') !== -1) {
        $(element).removeClass('error').addClass('warning');
    	} else if (error.text().indexOf('danger') !== -1) {
      	$(element).removeClass('error').addClass('danger');
      }
    },
    }
  });

});
```

rip747:
collect raised error in an object fixes #1925
All errors that are raised will be collected into a raisedErrors object on the base validator object (which can be accessed from an).  The id of the element will be used as the key with an empty object as the value. Each corresponding error with have the method name (such as "required") as the key with the error message as the value.

So for instance... if you have the following rules and messages:

rules: {
    email: { required: true, email: true }
},
messages: {
    email: {required: "Enter your email address", email: "A valid email address is required"}
},

would have a rasiedErrors object of:

this.rasiedErrors["email"]["required"] = "Enter your email address"
this.rasiedErrors["email"]["email"] = "A valid email address is required"

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -372,6 +372,7 @@ $.extend( $.validator, {
 			this.pending = {};
 			this.invalid = {};
 			this.reset();
+			this.raisedErrors = {};
 
 			var groups = ( this.groups = {} ),
 				rules;
@@ -856,15 +857,24 @@ $.extend( $.validator, {
 
 		formatAndAdd: function( element, rule ) {
 			var message = this.defaultMessage( element, rule );
+			var id = element.id;
+			var method = rule.method;
 
 			this.errorList.push( {
 				message: message,
 				element: element,
-				method: rule.method
+				method: method,
+				id: id
 			} );
-
+			
 			this.errorMap[ element.name ] = message;
 			this.submitted[ element.name ] = message;
+			
+			if(!this.raisedErrors.hasOwnProperty(id)){
+				this.raisedErrors[id] = {};
+			}
+			this.raisedErrors[id][method] = message;
+			
 		},
 
 		addWrapper: function( toToggle ) {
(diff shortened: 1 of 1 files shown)


staabm:
this new list needs to be cleared sometime..

do we really need this new list of errors or can we re-use one of the existing error lists?

rip747:
@staabm  Thinking about this more... I agree, I think I was getting a little carried away with this.

PR #1994, opened 2017-04-18
Update cifES.js

manospasj:
remove unreachable return


src/additional/cifES.js
```diff
@@ -108,6 +108,4 @@ $.validator.addMethod( "cifES", function( value ) {
 		return control === control_digit || control === control_letter;
 	}
 
-	return false;
```
staabm: could you remove the else block and move the `return` statement from the else-block to this line here?

manospasj: Done

Arkni:

PR #2010, opened 2017-05-27
Core: Added :and expression

mikaelkaron:
It's often desirable to require a field only if conditions a _and_ b are fulfilled. This expression lets us do this in a simple manner using a custom expression.

In this case we want `confirm` to be required if the value of `first` is `"John"` or `"Jane"` and the value of `last` is `"Doe"`

```html
<input id="first">
<input id="last">
<input id="confirm" type="checkbox" data-required="#last[value='Doe']:and(#first[value='John'], #first[value='Jane'])" />
```

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -197,6 +197,10 @@ $.extend( $.fn, {
 // Custom selectors
 $.extend( $.expr.pseudos || $.expr[ ":" ], {		// '|| $.expr[ ":" ]' here enables backwards compatibility to jQuery 1.7. Can be removed when dropping jQ 1.7.x support
 
+	and: function( a, b, c ) {
+		return $.find( c[ 3 ] ).length !== 0;
+	},
+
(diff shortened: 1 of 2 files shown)


mikaelkaron:
Could also be a good idea to add this if you want to validate against values:

```javascript
$("input").on("change", function($event) {
  var $target = $($event.target);
  $target.attr("value", $target.val());
});
```

staabm:
I dont think that it make sense to allow such complex validation rules via markup. Highly customized or otherwise custom domain logic should be implemented in plain JS not via any kind of markup expressions

staabm:
Will close for now. Feel free to ask further questions regarding this topic here. We can re-open if required.

mikaelkaron:
No worries. Published as a separate jQuery plugin as [mu-jquery-and](/mu-lib/mu-jquery-and) if anyone needs it.

PR #2011, opened 2017-05-29
Update core.js : Try to find element by id in the last case

gotzsys:
If the input has no name, use attribute id.

This functionality is very important to valid credit card forms in PCI DSS standards when input name is not defined, to not send any card data to the server.

https://www.pcisecuritystandards.org/security_standards/


staabm:
Couldnt we just add a new additional method which handles this exact case instead of changing a core assumption for all validators which is only required for a certain case..?

Issue #1841, opened 2016-08-17
Aria-required being added to unsupported roles/elements


sjlawton:
### Subject of the issue

The aria-required attribute is being added to radio buttons if they are required, when it is only allowed for a [specific set of roles ](https://www.w3.org/TR/wai-aria/states_and_properties#aria-required). This causes a ["This element has an unsupported ARIA attribute"](https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_10) error in validators like Google's Developer Accessibility Tools.
### Your environment

Jquery validate 1.15.0
Chrome 52.0.2743.116
### Steps to reproduce

To reproduce, validate a form with a radio button with the required attribute.
https://jsfiddle.net/Steven_Lawton/dr98hqL5/13/
### Expected behaviour

It is an acceptable use of the required attribute to put "required" on each radio button without a radiogroup above it. (https://www.w3.org/TR/html5/forms.html#the-required-attribute). However, it is not acceptable to use the aria-required attribute in this way (https://www.w3.org/TR/wai-aria/states_and_properties#aria-required). Instead, the aria-required attribute should be applied to a parent element with the radiogroup role. This also is true with tree items & their parent trees.

When an element that doesn't support aria-required is validated and has the required attribute, the aria-required attribute should not be added to it. 

If a radiobutton or tree item element has the required attribute, and its parent has the corresponding radiogroup/tree aria-role attribute, then the parent could receive the "aria-required" attribute rather than the 
### Actual behaviour

When any element with the required attribute is validated, the aria-required attribute is added to it.

see line 170-175

> ```
>   if ( data.required ) {
>       param = data.required;
>       delete data.required;
>       data = $.extend( { required: param }, data );
>       $( element ).attr( "aria-required", "true" );
>   }
> ```

staabm:
I agree this is a problem which needs fixing.

PR #2012, opened 2017-06-01
Core: removed aria-required attribute

alexbokii:
According to the latest documentation, 'aria-required' is a widget attribute and should be used only with particular roles. We don’t need to use it with native elements and when ‘required’ attribute exist. At the moment all instances of 'aria-required' attribute cause an error "This element has an unsupported ARIA attribute" in Google Chrome accessibility validator.
https://www.w3.org/TR/wai-aria/states_and_properties#aria-required

This addresses #1841.


Arkni:
@alexbokii 
Does the same apply to `contenteditable` elements like `div`s ?
What should happen in that case ?

alexbokii:
@Arkni No, it's different when we use divs - we need to apply aria-required attribute if the div tag is used instead of the input tag and user input is required. But the div should have a role attribute that points to the nature of control - it can be a "textbox", "combobox", etc.
So, according to this doc - https://www.w3.org/TR/wai-aria/states_and_properties#attrs_widgets, if we want to use aria-required with divs we must also add a role:
`<div contenteditable="true" role="textbox" aria-required="true">`.

staabm:
IIRC we support only real input types. When people use custom markup instead of native input elements they should apply the aria roles for them.

IMHO we should stop adding aria attributes.

👍

Arkni:
@alexbokii 

alexbokii:
@Arkni happy to help :)

PR #2019, opened 2017-06-25
stopRequest() doesn't pass on the submit button value (based on #1596)

Arkni:
Fixed fiddle: https://jsfiddle.net/Arkni/sofq4qbw/9/

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -26,9 +26,9 @@ $.extend( $.fn, {
 		if ( validator.settings.onsubmit ) {
 
 			this.on( "click.validate", ":submit", function( event ) {
-				if ( validator.settings.submitHandler ) {
-					validator.submitButton = event.currentTarget;
-				}
+				// Track the used submit button to properly handle scripted
+				// submits later.
+				validator.submitButton = event.target;
 
 				// Allow suppressing validation by adding a cancel class to the submit button
 				if ( $( this ).hasClass( "cancel" ) ) {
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -1083,6 +1089,15 @@ $.extend( $.validator, {
 			$( element ).removeClass( this.settings.pendingClass );
 			if ( valid && this.pendingRequest === 0 && this.formSubmitted && this.form() ) {
 				$( this.currentForm ).submit();
+
+				// Remove the hidden input that was used as a replacement for the
+				// missing submit button. The hidden input is added by `handle()`
+				// to ensure that the value of the used submit button is passed on
+				// for scripted submits triggered by this method
+				if ( this.submitButton ) {
+					$( "input:hidden[name='" + this.submitButton.name + "']" ).remove();
```
staabm: Can we scope this to the current form?

Arkni: Fixed!

staabm:
I guess we need a test (and maybe a fix) that the submit button is also posted when the form gets submitted via ENTER key

Arkni:
I'm afraid we can't simulate the default behaviour of the ENTER key, we can only simulate the keyboard event (`keypress`, `keyup` & `keydown`), other than that, it needs user interaction.

Also, if you press the ENTER key on the input of the above fiddle it will send the form alongside the first submit button. So, it doesn't need fixing.

staabm:
> I'm afraid we can't simulate the default behaviour of the ENTER key, we can only simulate the keyboard event (keypress, keyup & keydown), other than that, it needs user interaction.

hmm wouldnt a trigger of a native form submit event be enough for testing?

staabm:
> Also, if you press the ENTER key on the input of the above fiddle it will send the form alongside the first submit button. So, it doesn't need fixing.

perfect!

Arkni:
> hmm wouldnt a trigger of a native form submit event be enough for testing?

Nope. Doing so will only submit the form without the value of the submit button.

PR #2028, opened 2017-07-09
Readme: Add note about trimming whitespaces inside required method

Arkni:
Ref https://github.com/jquery-validation/jquery-validation/issues/2027#issuecomment-313874350


staabm:
Awesome, thx

PR #2032, opened 2017-07-14
Core: Add sanity checks to dateISO method

### LINKED ISSUE #1861: date/dateISO do not check days of month correctly
as of here: https://jqueryvalidation.org/dateISO-method/

both of these dates don't validate correctly, they should fail

2011-11-31
2011-02-29

normally i would use this which checks the data parsed from the input string matches the supplied date

``` javascript
var d = new Date(Date.parse(str))
return str === (d.getFullYear() + '-' + (d.getMonth()+1) + '-' + d.getDate());
```

since 2011-02-29 would parse as 2011-03-01, then this would fail properly

please advise

thanks
J

RobJohnston:
Updated regex for dateISO to match the ISO 8601 standard.  Added checks for leap years and for months that only have 30 days.  

This is a redo of PR #1528 due to mangled repo. 

Fixes #1861.

I don't expect it to be committed yet for the reasons outlined in https://github.com/jquery-validation/jquery-validation/pull/1528#issuecomment-150042586.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1382,8 +1382,37 @@ $.extend( $.validator, {
 		},
 
 		// https://jqueryvalidation.org/dateISO-method/
-		dateISO: function( value, element ) {
-			return this.optional( element ) || /^\d{4}[\/\-](0?[1-9]|1[012])[\/\-](0?[1-9]|[12][0-9]|3[01])$/.test( value );
+		dateISO: function (value, element) {
+			if (this.optional(element)) {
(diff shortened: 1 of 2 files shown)


nschonni:
FYI, came accross this regex that deals with the week, month, and time variants http://www.pelagodesign.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/
That still doesn't cover the leap parts you added though

PR #2033, opened 2017-07-14
Localization: Add French translation for notEqualTo method.

RobJohnston:
This localization seems to be missing in every language.


Arkni:

PR #2044, opened 2017-08-07
focusCleanup Validation Bug

bigcafe:
Valid class is added to all inputs on focus, including empty and invalid inputs.

If 'focusCleanup' is true, error class is successfully removed on focus as intended. However, valid class is also added to all inputs on focus, including empty and invalid inputs.


staabm:
Could you provide a jsfiddle which demonstrates how this change fixes a real world problem?

Issue #2049, opened 2017-08-23
Normalizer can't return array


YodaDaCoda:
The normaliser function requires a string to be returned - an array, such as you get for a multiselect, is not permitted. This seems wrong to me.

I initially started using the normalizer to prevent blank strings from being permitted, but I'd like to be able to properly validate multiselect picklist values too.

### Your environment
* version of `jquery-validate`: 1.17.0

### Expected behaviour
I expect the validator to just use the value that's returned from the normalizer, not attempt to sanity-check it. That's my responsibility.

### Actual behaviour
A TypeError is thrown and validation stops.

Arkni:
TBH, I don't remember why we required it to return a `string`. Nevertheless, I'll take a look at it soon.

Also, should we allow the `normalizer` to return `undefinded` or `null` ? as not all methods are built to check for them first (including the `required` method).

YodaDaCoda:
I know that older versions of jQuery (e.g. 1.12.4) return `null` for the value of a `select[multiple]` when no options are selected. See fiddle: https://jsfiddle.net/YodaDaCoda/0zb5az6v/

Later versions of jQuery return an empty array, but I'm not sure exactly when that behaviour changed.

Given that the website suggests that this plugin supports jQuery back to 1.7.2 ("tested with 1.7.2, 1.8.3, 1.9.1, 1.11.1, 3.1.1") I suggest allowing the normaliser to return any value, and leaving it up to the author of the normalisation function to catch cases that may return undesirable values.

Arkni:
Thanks for the example and info.

That makes sense to me. I check the code of the required method and it handles both `null` and `array` value of the `SELECT` element, but will not handle a `null` or `undefined` value coming from the `normalizer` which need a fix.

I'll open a PR soon to fix those two issues.

PR #2053, opened 2017-09-02
Core: guard against null & undefined values in required method

Arkni:
Ref #2049

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1349,7 +1349,7 @@ $.extend( $.validator, {
 			if ( this.checkable( element ) ) {
 				return this.getLength( value, element ) > 0;
 			}
-			return value.length > 0;
+			return value !== undefined && value !== null && value.length > 0;
 		},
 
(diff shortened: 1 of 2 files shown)


Arkni:
I don't think there is a way to test it without using the normalizer. But the normalizer will not return `undefined` or `null` until #2054 was merged.

I will update the tests of the other PR and merge it and then update this one to use the normalizer.

For now, I added a description to the commit describing the needs for this fix.

staabm:
Merged the other pr, we can add the tests here

Arkni:
We are good to go.

Arkni:
I'm going to merge this for now.
Thanks for the review and for merging the other PR.

staabm:
Awesome that our normalizer layer also turned out to be a good value-mocking test utility 😸

Arkni:
Yes, and also I see (from Stackoverflow and here) that a lot of folks are using it and found it useful too.

Thanks :)

PR #2054, opened 2017-09-02
Core: Allow the normalizer to return any value

Arkni:
Needs #2053 to be merged before in order for tests to pass


staabm:
Merged the other pr, we can add the tests here

PR #2057, opened 2017-09-03
Build: Include minified version of additional methods in npm package

Arkni:


Arkni:
@staabm 

Should we also add the `localization` folder to tags? In other words, adding it to this list: https://github.com/jquery-validation/jquery-validation/blob/5482dfd95ff01bb89a4fe5171f67c63d90c12198/build/release.js#L39-L44

staabm:
Since nobody requested localizations in the npm package I would leave them out for now.

Whats your opinion?

Arkni:
The localization folder is already distributed in the npm package, see https://github.com/jquery-validation/jquery-validation/blob/5482dfd95ff01bb89a4fe5171f67c63d90c12198/package.json#L30-L31

I'm talking about the dist folder committed to GitHub as part of the released tag. See [the dist folder](https://github.com/jquery-validation/jquery-validation/tree/1.17.0/dist) of the release `1.17.0` as an example.

staabm:
oh I see. please add it then.

build/release.js
```diff
@@ -37,6 +37,7 @@ Release.define({
 	generateArtifacts: function( done ) {
 		Release.exec( "grunt release", "Grunt command failed" );
 		done([
```
staabm: sorry for beeing late to the party:

please add a note like `keeps this is of files in sync with package.json "files"-key`

Arkni: Done!

staabm:
thx!

PR #2061, opened 2017-09-09
Core: Add greaterThan and lessThan methods.

### LINKED ISSUE #2030: Add greaterThan and lessThan methods.
Similar to the [equalTo method,](https://jqueryvalidation.org/equalTo-method/) and the [notEqualTo method](https://github.com/jquery-validation/jquery-validation/blob/master/src/additional/notEqualTo.js), would there be interest in having  `greatherThan `and `lessThan `methods?  Additionally, `greaterThanEqual `and `lessThanEqual `for completeness?  I'm thinking that these can be numeric <ins>and date</ins> comparisons only.

The situation would be two text boxes, one being a "Start" value and the other an "End" value.  The [range method](https://jqueryvalidation.org/range-method/) could be used here as well, but the "End" value must be still be greater than the "Start" value.

RobJohnston:
Fixes #2030.

Tests are limited to integers, but I think that other data types would work too.

### DIFF
--- a/src/additional/greaterThan.js
+++ b/src/additional/greaterThan.js
@@ -1,0 +1,11 @@
+$.validator.addMethod( "greaterThan", function( value, element, param ) {
+    var target = $( param );
+
+    if ( this.settings.onfocusout && target.not( ".validate-greaterThan-blur" ).length ) {
+        target.addClass( "validate-greaterThan-blur" ).on( "blur.validate-greaterThan", function() {
+            $( element ).valid();
+        } );
+    }
+
+    return value > target.val();
+}, "Please enter a greater value." );
--- a/src/additional/greaterThanEqual.js
+++ b/src/additional/greaterThanEqual.js
@@ -1,0 +1,11 @@
+$.validator.addMethod( "greaterThanEqual", function( value, element, param ) {
+    var target = $( param );
+
+    if ( this.settings.onfocusout && target.not( ".validate-greaterThanEqual-blur" ).length ) {
+        target.addClass( "validate-greaterThanEqual-blur" ).on( "blur.validate-greaterThanEqual", function() {
+            $( element ).valid();
+        } );
+    }
+
+    return value >= target.val();
+}, "Please enter a greater value." );
--- a/src/additional/lessThan.js
+++ b/src/additional/lessThan.js
@@ -1,0 +1,11 @@
+$.validator.addMethod( "lessThan", function( value, element, param ) {
+    var target = $( param );
+
+    if ( this.settings.onfocusout && target.not( ".validate-lessThan-blur" ).length ) {
+        target.addClass( "validate-lessThan-blur" ).on( "blur.validate-lessThan", function() {
+            $( element ).valid();
+        } );
+    }
+
+    return value < target.val();
+}, "Please enter a lesser value." );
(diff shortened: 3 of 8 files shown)


Arkni:
Thanks for the PR.

Looks good to me.

Can you also unbind the event and remove the class name in `validator#destroy()` as done for `equalTo` method? https://github.com/jquery-validation/jquery-validation/blob/539fa4d1426ed3d6ab6955865040660c56d3205e/src/core.js#L1121-L1123

As for testing the above request, you can do the same as the PR #1709

RobJohnston:
Not sure I fully grok the tests... is it expected that the following small change should still pass?

```js
diff --git a/test/test.js b/test/test.js
index 1ee1d53..4138087 100644
--- a/test/test.js
+++ b/test/test.js
@@ -2453,8 +2453,8 @@ QUnit.test( "destroy()", function( assert ) {
 
        validate.destroy();
        assert.strictEqual( form.data( "validator" ), undefined );
-       assert.equal( $( "#x1", form ).hasClass( "validate-equalTo-blur" ), false, "The blur event should be unbound from this element" );
-       assert.equal( $( "#x2", form ).hasClass( "validate-equalTo-blur" ), false, "The blur event should be unbound from this element" );
+       assert.equal( $( "#x1", form ).hasClass( "validate-equalToxxx-blur" ), false, "The blur event should be unbound from this element" );
+       assert.equal( $( "#x2", form ).hasClass( "validate-equalToxxx-blur" ), false, "The blur event should be unbound from this element" );
 } );
 
 QUnit.test( "#1618: Errorlist containing more errors than it should", function( assert ) {
```

RobJohnston:
@Arkni So did I get it, or do I need to do some more studying of this repo?

src/additional/greaterThan.js
```diff
@@ -0,0 +1,11 @@
+$.validator.addMethod( "greaterThan", function( value, element, param ) {
+    var target = $( param );
+
+    if ( this.settings.onfocusout && target.not( ".validate-greaterThan-blur" ).length ) {
```
staabm: Why do we need this focus handling?

RobJohnston: My motivation was to follow the pattern of the [equalTo](https://github.com/jquery-validation/jquery-validation/blob/master/src/core.js#L1466) method, and this was introduced there two years ago in [Core: Bind the `blur` event just once in `equalTo` rule ](https://github.com/jquery-validation/jquery-validation/commit/ef223608a31e487001006ea03776024cf46622cd).

Arkni: Just a clarification, the focus handling was added +5 years ago in this commit https://github.com/jquery-validation/jquery-validation/commit/aca144b5eba1f1947b311819e9eea8e0f176f448#diff-a692042a498ec7e201f38855d169e7b1R1163. The only thing that was added in the commit linked by @RobJohnston is binding the blur event one time.

staabm:
@RobJohnston do you have a actual use case for this newly added methods?

I am fine with adding them in case there is a need for it.
Only thing I dont like is the few lines added in core.js, but ok in case of a use case..

RobJohnston:
Yes I do.  I have an application where people must record the inventory that they use, and each item in inventory has a serial number.  At the end of the day, they have a range of consumed inventory, and this validation would ensure that the start serial number is less than the end serial number (or end is greater than start).  If they only consumed one piece of inventory, then the end serial number has to be greater than or equal to the start serial number.

I thought it would be an easy PR to accept due to the previous discussion in https://github.com/jquery-validation/jquery-validation/pull/458#issuecomment-14171064.

staabm:
Thx for your detailed feedback

PR #2063, opened 2017-09-12
Setting: Adding ability to override the default element tag list.

WalterWeidner:
This override allows elements implementing an input interface to be validated

#### Description
We have been using jQuery validation for several years and it has been very useful for our organization. Unfortunately, one area that it falls short is where it doesn't support custom elements that behave like native inputs. There is no way (without overriding methods) that we know of to validate any input type other than those that are hard coded in the `elements()` prototype method. This Pull Request attempts to provide a simple method for customizing what tags the `element()` function looks for.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -265,6 +265,7 @@ $.extend( $.validator, {
 		errorLabelContainer: $( [] ),
 		onsubmit: true,
 		ignore: ":hidden",
+		inputSelector: "input, select, textarea, [contenteditable]",
 		ignoreTitle: false,
 		onfocusin: function( element ) {
 			this.lastActive = element;
@@ -616,7 +617,7 @@ $.extend( $.validator, {
 
 			// Select all valid inputs inside the form (no submit or reset buttons)
 			return $( this.currentForm )
-			.find( "input, select, textarea, [contenteditable]" )
+			.find( validator.settings.inputSelector )
 			.not( ":submit, :reset, :image, :disabled" )
 			.not( this.settings.ignore )
 			.filter( function() {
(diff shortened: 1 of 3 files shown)


staabm:
With custom element you mean actually a webcomponent?

src/core.js
```diff
@@ -616,7 +617,7 @@ $.extend( $.validator, {
 
 			// Select all valid inputs inside the form (no submit or reset buttons)
 			return $( this.currentForm )
-			.find( "input, select, textarea, [contenteditable]" )
+			.find( validator.settings.inputSelector )
```
staabm: validator -> this

src/core.js
```diff
@@ -265,6 +265,7 @@ $.extend( $.validator, {
 		errorLabelContainer: $( [] ),
 		onsubmit: true,
 		ignore: ":hidden",
+		inputSelector: "input, select, textarea, [contenteditable]",
```
staabm: Other settings dont use the suffix `Selector` we shouldnt use it either.

staabm: Not sure yet what a good name should be for this setting.

Arkni: What about `targets`?
It aligns well with the internal method `validationTargetFor(...)`.

Arkni:
Over all, it looks good to me. 

One small thing that come to my mind while reading the description & the patch. How can the plugin attach events (`click`, `keyup`, ...) to the custom elements as it does for standard ones? The chunk of code that attaches the event is: https://github.com/jquery-validation/jquery-validation/blob/c62e0d91bcf05bcda22e5cfa4068fcf6d4a21f2b/src/core.js#L409-L418

What do you propose to fix that?

Arkni:
Hi @WalterWeidner 

Any updates regarding this PR?

Thanks!

PR #2070, opened 2017-09-13
Added Danish(DK) Phone number

BenjaminHoegh:
Matches DK phone number format
 
allows '-' or ' ' as a separator and allows parens around area code
some people may want to put a '+45' in front of their number
 
 - +45 21-77-21-77
 - +45 21 77 21 77
 - +45 21772177
 - +4521772177
 - 21 77 21 77
 - 21-77-21-77
 - 2177 2177
 - 2177-2177
 - 21772177
 
 but not
 - +1 3243121212
 - +45 112
 - +452143284233
 - 2143284233

### DIFF
--- a/src/additional/phoneDK.js
+++ b/src/additional/phoneDK.js
@@ -1,0 +1,27 @@
+/**
+ * Matches DK phone number format
+ *
+ * allows '-' or ' ' as a separator and allows parens around area code
+ * some people may want to put a '+45' in front of their number
+ *
+ * +45 21-77-21-77
+ * +45 21 77 21 77
+ * +45 21772177
+ * +4521772177
+ * 21 77 21 77
+ * 21-77-21-77
+ * 2177 2177
+ * 2177-2177
+ * 21772177
+ *
+ * but not
+ * +1 3243121212
+ * +45 112
+ * +452143284233
+ * 2143284233
+ */
+$.validator.addMethod( "phoneDK", function( phone_number, element ) {
+	phone_number = phone_number.replace( /\s+/g, "" );
+	return this.optional( element ) || phone_number.length > 9 &&
+		phone_number.match( /^(\+45( )?)?([0-9]{2})((-| )?[0-9]{2})((-| )?[0-9]{2})((-| )?[0-9]{2})$/ );
+}, "Please specify a valid phone number" );


staabm:
Please add tests

HawkonDK:
@staabm what do you mean ??? screenshots?

staabm:
I mean unit tests, see the tests folder. I also realized we dont have tests for the other languages, so this is not a must have.

@Arkni could you add some example tests for phoneUs so people see a example?

src/additional/phoneDK.js
```diff
@@ -0,0 +1,27 @@
...
+ * but not
+ * +1 3243121212
+ * +45 112
+ * +452143284233
+ * 2143284233
+ */
+$.validator.addMethod( "phoneDK", function( phone_number, element ) {
+	phone_number = phone_number.replace( /\s+/g, "" );
+	return this.optional( element ) || phone_number.length > 9 &&
+		phone_number.match( /^(\+45( )?)?([0-9]{2})((-| )?[0-9]{2})((-| )?[0-9]{2})((-| )?[0-9]{2})$/ );
```
Arkni: Please use `regex.test(string)` instead of `string.match(regex)`. We only need to know whether the phone number matches the regex or not. We don't need to retrieve the matched parts.

Also, please use non-capturing instead of capturing groups [1]

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#grouping-back-references

Arkni:
@staabm 
I will see if I can find a moment to work on them tomorrow.

Tests should not block us from merging this PR though, as @HawkonDK can add them in another PR.

src/additional/phoneDK.js
HawkonDK: okay thanks, will look on it later this week

staabm: @Arkni can this regex codestyle be enforced via jshint?

Arkni:
@staabm 
Just did a quick read to the test files and I found that tests for `phoneUS` are already added (`phoneUK` tests are there too). See https://github.com/jquery-validation/jquery-validation/blob/38cb9b9c8c135d9737d56f4a19ee48ca9a2d888b/test/methods.js#L779-L789

src/additional/phoneDK.js
Arkni: None that I am aware of.

staabm:
@Arkni  awesome, we should separate them into a „additional“ folder and equal named files like the impl the

Arkni:
Hi @HawkonDK 
Are you willing to continue working on this?

Thanks!

PR #2071, opened 2017-09-13
Added Postal code DK

BenjaminHoegh:
This accept all danish(DK) zip codes (1000 to 9999).
 also work for the danish kingdoms Greenland(use same digital zip codes).
 Not work for the danish kingdom Faroe Islands after they use 3 digi zip codes.

Note: Greenland and Faroe Islands is a part of the Danish Kingdom and not the country it self.

### DIFF
--- a/src/additional/postalcodeDK.js
+++ b/src/additional/postalcodeDK.js
@@ -1,0 +1,12 @@
+/* Matches Danish postcode */
+$.validator.addMethod( "postalcodeDK", function( value, element ) {
+	return this.optional( element ) || /^[0-9]{4}$/.test( value );
+}, "Please specify a valid postal code" );
+
+/*
+ * This accept all danish(DK) zip codes (1000 to 9999)
+ * also work for the danish kingdoms Greenland.
+ * Not work for the danish kingdom Faroe Islands after they use 3 digi zip codes.
+ *
+ * Note: Greenland and Faroe Islands is a part of the Danish Kingdom and not the country it self.
+*/


src/additional/postalcodeDK.js
```diff
@@ -0,0 +1,12 @@
+/* Matches Danish postcode */
+$.validator.addMethod( "postalcodeDK", function( value, element ) {
+	return this.optional( element ) || /^[0-9]{4}$/.test( value );
+}, "Please specify a valid postal code" );
+
+/*
+ * This accept all danish(DK) zip codes (1000 to 9999)
```
staabm: Please move this comment to the top

src/additional/postalcodeDK.js
```diff
@@ -0,0 +1,12 @@
+/* Matches Danish postcode */
```
staabm: Replace with the extensive comment from below

staabm:
Please add tests

RobJohnston:
Optionally, you _could_ chage the regex to `^[0-9]{3,4}$` to get it to work with the Faroe Islands as well.

Arkni:
Hi @HawkonDK 
Are you willing to continue working on this?

Thanks!

src/additional/postalcodeDK.js
BenjaminHoegh: Oh! why did I not see that? a bigger question is why GitHub didn't notify me about your feedback

BenjaminHoegh:
Yes @Arkni Github for some reason didn't notify me about your comments

PR #2073, opened 2017-09-16
Fix typos in messages_pt_BR.js

CRC-Mismatch:
I found some typos in the pt_BR localization messages. Nothing big, but it's worth correcting since I'm going to use the library in a production environment.


staabm:
Thx

PR #2074, opened 2017-09-19
Core: Add multiple attribute to email method.

### LINKED ISSUE #1722: email input type and the multiple attribute
HTML5 has a [multiple attribute](https://www.w3.org/TR/2012/WD-html5-20121025/common-input-element-attributes.html#attr-input-multiple) that can be specified for e-mail addresses as well as files.  For e-mail, it would be [a set of comma-separated strings, each of which is a valid email address](https://www.w3.org/TR/html-markup/input.email.html#input.attrs.multiple).  

When applying the multiple attribute to the existing [jQuery Validation Plugin Demo](http://jqueryvalidation.org/files/demo/), more than one e-mail address will not validate.

There are solutions to this (e.g., http://stackoverflow.com/questions/10352221/jquery-validation-plugin-multiple-email-addresses), but ideally, one should be able to write the following without using addMethod:

``` html
<input id="email1" name="email1" type="email" multiple>
```

RobJohnston:
Using @Sven-Mad's method. Fixes #1722.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1357,13 +1357,28 @@ $.extend( $.validator, {
 		},
 
 		// https://jqueryvalidation.org/email-method/
-		email: function( value, element ) {
+		email: function( value, element, no_multiple ) {
 
 			// From https://html.spec.whatwg.org/multipage/forms.html#valid-e-mail-address
 			// Retrieved 2014-01-14
 			// If you have a problem with this implementation, report a bug against the above spec
 			// Or use custom methods to implement your own email validation
-			return this.optional( element ) || /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/.test( value );
+			if ( this.optional( element ) ) {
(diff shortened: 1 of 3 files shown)


RobJohnston:
I think this PR still has merit.

Arkni:
Hi @RobJohnston,

Sorry for taking so long to review this PR. I only have one concern about this implementation. That `no_multiple` argument will always be `true` whether you are setting the rule using inline HTML or in JS.

What I would like to suggest is, we add support for passing an object to `email` method in JS like the following:
```js
$("form").validate({
  rules: {
    field: {
      email: {
        multiple: true | false
      }
    }
  }
});
```

The `no_multiple` parameter will need to be changed to `param` (or another convenient name).

RobJohnston:
Hi @Arkni .  It's now my time to apologize to you for the delay!  I've been away from front-end development for some time.  I didn't quite understand your comment when I first read it and not sure if I do now.

From the W3, the [multiple attribute](https://www.w3.org/TR/2012/WD-html5-20121025/common-input-element-attributes.html#attr-input-multiple) is what they call a [boolean attribute](https://www.w3.org/TR/2012/WD-html5-20121025/common-microsyntaxes.html#boolean-attribute), but that doesn't mean that it can be set to "true" or "false".  It must be `multiple="multiple"` or simply just 'multiple'.  Does that change your suggestion where you wrote:
```js
multiple: true | false
```

maysam (approved):

PR #2075, opened 2017-09-19
Localization: Add missing format method.

### LINKED ISSUE #2065: Pattern return english error message. Other language set
Pattern return "Invalid format" in English even if you have chose another language

RobJohnston:
I searched all localization files for "{0}" to see if I could find any without a call to the `format `method.  Found two.

Partially fixes #2065 (the French part, identified by @HawkonDK).  Danish translations were added in https://github.com/jquery-validation/jquery-validation/pull/2067, so that issue can be closed with this PR.


Arkni:

PR #2076, opened 2017-09-19
Shield.io release badget

HawkonDK:
Added Shield.io release badget and a tab error in one of the code examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 [jQuery Validation Plugin](https://jqueryvalidation.org/) - Form validation made easy
 ================================
 
+[![release](https://img.shields.io/github/release/jquery-validation/jquery-validation.svg)](https://github.com/jquery-validation/jquery-validation/releases/latest)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,7 @@
 [jQuery Validation Plugin](https://jqueryvalidation.org/) - Form validation made easy
 ================================
 
+[![release](https://img.shields.io/github/release/jquery-validation/jquery-validation.svg)](https://github.com/jquery-validation/jquery-validation/releases/latest)
```
staabm: @Arkni wdyt...? I dont think we need this badge?

HawkonDK: fair enough just nice to see the version number so you don't need to go under releases to see if there are any new

Arkni: It's a good addition. I've seen a number of npm packages doing the same. It gives the visitor a glimpse of the current version without the need to check the changelog or the releases tab (as @HawkonDK said)

staabm:
Thx

PR #2082, opened 2017-09-24
Build: Generate sub-resource integrity hashes of the distribution files

Arkni:
With this, running `grunt release` will generate ~~3 new files~~ 1 file in the `dist` folder:
* ~~`sri-core.json`: contains the hashes of `jquery-validate.js` and `jquery-validate.min.js`~~
* ~~`sri-additional.json`: contains the hashes of `additional-methods.js` and `additional-methods.min.js`~~
* ~~`sri-localization.json`: contains the hashes of all JS files inside `localization` folder~~
* `jquery-validation-sri.json`: contains the hashes of all JS files inside `dist` folder

I can add those files to the list of [distribution files][1] in case we want to commit them to GitHub.

[1]: https://github.com/jquery-validation/jquery-validation/blob/b70fcc20f1bcdcd63422686595a48950326309f1/build/release.js#L40-L46

//c @staabm @vanillajonathan

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -171,6 +171,27 @@ grunt.initConfig( {
 				}
 			]
 		}
+	},
+
+	// Generate the sub-resource integrity hashes of the distribution files
+	sri: {
+		options: {
+			algorithms: [ "sha256", "sha384", "sha512" ],
+
+			// The target json file
+			dest: "dist/jquery-validation-sri.json",
+
+			// Stringify the JSON output in a pretty format
+			pretty: true
+		},
+
+		all: {
+			src: [
+				"dist/jquery.validate.{min.js,js}",
+				"dist/additional-methods.{min.js,js}",
+				"dist/localization/*.js"
+			]
+		}
 	}
 } );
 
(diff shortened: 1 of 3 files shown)


staabm:
Looks cool. Did we discuss using a single sri file vs the 3 you added?
Need to test it locally before merge.

Arkni:
No, we didn't. I just felt the need for separating and categorizing the hashes by component.

staabm:
I guess it would be easier for tool driven consumption when all hashes are contained in a single file.

Arkni:
Make sense. I will update it this evening.

Arkni:
@staabm 
Updated per suggestion!

Arkni:
Hi @staabm 

any comment on this PR?
Should I add those files to the list of [distribution files][1] in case we want to commit them to GitHub?

[1]: https://github.com/jquery-validation/jquery-validation/blob/b70fcc20f1bcdcd63422686595a48950326309f1/build/release.js#L40-L46

staabm:
I love this change.

> Should I add those files to the list of distribution files in case we want to commit them to GitHub?

I guess we should.

vanillajonathan:
I would like to see the hash on the website, next to the download link.

Arkni:
> we dont have this dist folder in our codebase on github.. I dont get what you mean here?

We publish the `dist` folder with every tagged release. For example, see https://github.com/jquery-validation/jquery-validation/tree/1.17.0
So, if someone navigates the tags, they will see the `dist` folder with its content (`additional-methods.js`, `additional-methods.min.js`, `jquery.validate.js`, `jquery.validate.min.js`). And I would like to add the sri file to be added to this list too.

> I would like to see the hash on the website, next to the download link.

We can provide a link to the file in our docs. The file will be hosted here on GitHub, so everyone can access it without going to the website.

staabm:
ok cool. please go for it then.

Arkni:
@staabm 

Updated the proposal to include the sri hashes file in tagged releases.

staabm:
Looks great, thanks

PR #2087, opened 2017-10-03
Additional: max size and number of files

RobJohnston:
This is a continuation of abandoned PR https://github.com/jquery-validation/jquery-validation/pull/1512, in that it validates the maximum size of a selected file.

Since the "file" type can contain the multiple attribute, another method was added to validate the maximum size of all selected files.  Finally, a third method was added to validate the maximum number of files.

### DIFF
--- a/src/additional/maxfiles.js
+++ b/src/additional/maxfiles.js
@@ -5,10 +5,8 @@ $.validator.addMethod( "maxfiles", function( value, element, param ) {
 	}
 
 	if ( $( element ).attr( "type" ) === "file" ) {
-		if ( element.files && element.files.length ) {
-			if ( element.files.length > param ) {
-				return false;
-			}
+		if ( element.files && element.files.length && element.files.length > param ) {
+			return false;
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


src/additional/maxfiles.js
```diff
@@ -0,0 +1,16 @@
+//Limit the number of files in a FileList.
+$.validator.addMethod( "maxfiles", function( value, element, param ) {
+	if ( this.optional( element ) ) {
+		return true;
+	}
+
+	if ( $( element ).attr( "type" ) === "file" ) {
+		if ( element.files && element.files.length ) {
+			if ( element.files.length > param ) {
```
Arkni: Can you please merge the above two `if` blocks?

src/additional/maxfiles.js
```diff
@@ -0,0 +1,16 @@
+//Limit the number of files in a FileList.
```
Arkni: Just for consistency with the other comments in this project, can you add a space before the comment?
The same goes for other comments bellow.

src/additional/maxfiles.js
RobJohnston: Which two?  There are three of them.  Lines 8 and 9, I presume?

Arkni: Yes, 8 and 9.

RobJohnston:
@Arkni  Done.

Arkni:

PR #2098, opened 2017-10-27
Additional: allow N11 exchange for non-geo US phone. Fixes #1695

### LINKED ISSUE #1695: Adding support for a 1 digit in 3rd position of CO code for non-geographical area codes in PhoneUS validator
As discussed in a previously closed issue, non-geographical US area codes can have a 1 in the third position of the CO code. The current phoneUS validator does not support this.

For example:
888-511-xxxx (replacing xxxx with some real 4 digit number) is a valid US phone number but not accepted by the validator.

infused:
Adds support to the phoneUS validator for a 1 in the third position of the CO code to support non-geographical US area codes. 

For example:
800-811-xxxx (replacing xxxx with some real 4 digit number) is a valid US phone number but not currently accepted by the validator.


staabm:
thx

PR #2112, opened 2017-12-01
Translation: fix meaning in Bulgarian sentence

user890104:
The "maxlength" message has wrong meaning in Bulgarian. The current meaning is:
Please enter more than {0} characters.
The correct meaning should be:
Please enter no more than {0} characters.
I could use "less" (по-малко) to mean the same thing, but it does not sound natural in this context.


staabm:
thx

PR #2115, opened 2017-12-04
Update German language file

Hawkeye1909:
Added translation for the missing default methods and all additional ones.


staabm:
thx

PR #2138, opened 2018-02-15
Core: Add deprecation warning to 'date' method

Arkni:


src/core.js
```diff
@@ -1386,6 +1386,16 @@ $.extend( $.validator, {
 
 		// https://jqueryvalidation.org/date-method/
 		date: function( value, element ) {
+			if ( this.settings.debug && window.console ) {
```
staabm: should we warn only on the first occurence? Doing it all the time could get a perf problem?

Arkni: Good point. I'll update it shortly.

staabm:
Thx

PR #2142, opened 2018-02-21
Core: Use of jQuery attr instead of hasAttribute. (#2141)

### LINKED ISSUE #2141: Object doesn't support property or method 'hasAttribute' error appears on old IE browsers
I'm experiencing some issues with users that are working with old versions of IE (version 10 and below or 11 with enabled compatibility mode)
I have a really basic login page and using version 1.17.0 but this was happening also on previous versions.

The error appears when focusing out from a required field. The following error is displayed on the console:

**Object doesn't support property or method 'hasAttribute'**

The error is thrown on line 705

```
if ( element.hasAttribute( "contenteditable" ) ) {
	val = $element.text();
} else {
	val = $element.val();
}
```

Somehow on old versios of IE the element object doesn't contain that method.
I was able to fix this by changing that line to the jquery function hasOwnProperty instead.

` $element.hasOwnProperty( "contenteditable" )`

The same can be found on line 152 and the same error should appear there.

This is my validation setup:

```
$('.login-form').validate({
    errorElement: 'span', 
    errorClass: 'help-block', 
    focusInvalid: false, 
    rules: {
        UserName: {
            required: true
        },
        Password: {
            required: true
        }
    },
    messages: {
        UserName: {
            required: resources.IngresarUsuario
        },
        Password: {
            required: resources.IngresarPWD
        }
    },
    invalidHandler: function (event, validator) {
        $('.alert-danger', $('.login-form')).show();
    },
    highlight: function(element) { 
        $(element).closest('.form-group').addClass('has-error');
    },
    success: function (label) {
        $(label).closest('.form-group').removeClass('has-error');
        $(label).remove();
    },
    errorPlacement: function (error, element) {
        $(error).insertAfter($(element).closest('.input-icon'));
    },
});
```

lopezcjf:
Fixes not supported method hasAttribute on IE with enabled compatibility mode

Fixes #2141

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -131,7 +131,7 @@ $.extend( $.fn, {
 			return;
 		}
 
-		if ( !element.form && element.hasAttribute( "contenteditable" ) ) {
+		if ( !element.form && element.isContentEditable ) {
 			element.form = this.closest( "form" )[ 0 ];
 			element.name = this.attr( "name" );
 		}
@@ -393,7 +393,7 @@ $.extend( $.validator, {
 			function delegate( event ) {
 
 				// Set form expando on contenteditable
-				if ( !this.form && this.hasAttribute( "contenteditable" ) ) {
+				if ( !this.form && this.isContentEditable ) {
 					this.form = $( this ).closest( "form" )[ 0 ];
 					this.name = $( this ).attr( "name" );
 				}
@@ -626,7 +626,7 @@ $.extend( $.validator, {
 				}
 
 				// Set form expando on contenteditable
-				if ( this.hasAttribute( "contenteditable" ) ) {
+				if ( this.isContentEditable ) {
 					this.form = $( this ).closest( "form" )[ 0 ];
 					this.name = name;
 				}
@@ -684,7 +684,7 @@ $.extend( $.validator, {
 				return element.validity.badInput ? "NaN" : $element.val();
 			}
 
-			if ( element.hasAttribute( "contenteditable" ) ) {
+			if ( element.isContentEditable ) {
 				val = $element.text();
 			} else {
 				val = $element.val();


staabm:
thx for the PR!

how to reproduce the issue you are fixing? which IE version is affected?

lopezcjf:
Hi,
I reproduced this error by setting up a basic form with 2 required fields and using IE 11 with enabled compatibility mode (Go to settings -> Compatibility View Settings -> Add the website address, in my scenario is localhost)

Open the page
Click on the first field
Then click on the second field

When focusing out from the first field, the error appears.

lopezcjf:
This is my validation setup:

```
$('.login-form').validate({
    errorElement: 'span', 
    errorClass: 'help-block', 
    focusInvalid: false, 
    rules: {
        UserName: {
            required: true
        },
        Password: {
            required: true
        }
    },
    messages: {
        UserName: {
            required: resources.IngresarUsuario
        },
        Password: {
            required: resources.IngresarPWD
        }
    },
    invalidHandler: function (event, validator) { 
        $('.alert-danger', $('.login-form')).show();
    },
    highlight: function(element) { 
        $(element)
            .closest('.form-group').addClass('has-error'); 
    },
    success: function (label) {
        $(label).closest('.form-group').removeClass('has-error');
        $(label).remove();
    },
    errorPlacement: function (error, element) {
        $(error).insertAfter($(element).closest('.input-icon'));
    }
});
```

Arkni:
@staabm 

> how to reproduce the issue you are fixing? which IE version is affected?

[`hasAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute) is supported as of IE8. And because we still support down to jQuery 1.7.2 which in its own support IE 7, we have to replace all the 4 occurences of `hasAttribute` with their equivalent in jQuery using [`.attr()`](https://api.jquery.com/attr/).
That's the main reason why I added the `bug` label to the linked issue.

We definetly should drop older versions of jQuery in the next major release and stick to versions `>v1.10.0`, `>2.2.0` and `v3.0.0` or just support the latest stable version (`v3.x.y`)

staabm:
> We definetly should drop older versions of jQuery in the next major release and stick to versions >v1.10.0, >2.2.0 and v3.0.0 or just support the latest stable version (v3.x.y)

big 👍 .

we should gather some information which usefull BC breaks we should target for 2.0.

Arkni:
Hi @lopezcjf,

Sorry for taking so long to review this.

I did some research in order to have a global understanding of this issue. I found out that `contenteditable` is supported as of `IE6` and so we can easily replace the check with [`element.isContentEditable`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/isContentEditable) which returns `true` is the element is editable otherwise it returns `false`.
This will help us with the case where the element has no `contenteditable` attribute but it inherits the editable state from its editable parent ([`contenteditable = "inherit"`](https://www.w3.org/TR/html50/editing.html#attr-contenteditable)).

So, can you please update this PR to use `isContentEditable` property instead.

Arkni:
Hi @lopezcjf,

Just a freindly reminder.

Are you still willing to continue this PR? No worries if you don't have time for it, though.

Thanks!

Issue #2035, opened 2017-07-20
If input belongs to another form an error is thrown


dtritus:
### Subject of the issue
If form contains (as child node) input field with `form` attribute pointed to another form an error is thrown when you interact with input field (focus, input etc).

### Your environment
* jQuery Validation Plugin v1.16.0
* Chrome Version 59.0.3071.115

### Steps to reproduce
See the simple example: https://jsfiddle.net/hc06nmtj/1/
Click on second input and check console logs

### Expected behaviour
Ignore such fields, since they are not belong to validated form

### Actual behaviour
Error is thrown

staabm:
Agree, PR welcome. Thx for reporting.

Arkni:
I think this issue is related to this proposal #1970. We have two ways to handle this:
1. Ignore the element and warn the developer about it with the error message mentioned in the PR
2. Throw an error

I'm in favor with the first option as that what the developer expect to happen because they clearly declare that the element doesn't belong to the validated form.

staabm:
I am in favor of suggestion 1

PR #2147, opened 2018-03-04
Ignore elements that belong to other/nested forms

Arkni:


Arkni:
**Note to self**: Need to fix the case described in #2035 as this PR only ignores the elements when submitting the form or when validating it directly using either `$(form).valid()`, `validator.form()` or `validator.checkForm()`.

Arkni:
Fixed the issue mentionned in my previous comment.
This PR is good to go now.

PR #2149, opened 2018-03-06
Add Accessibility section to Readme

joe-watkins:
Adding Accessibility section to the readme that speaks to using `errorElement` in the `validate()` method.

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,25 @@ $("#myForm").validate({
 });
 ```
 
+## Accessibility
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,6 +73,27 @@ $("#myForm").validate({
 });
 ```
 
+## Accessibility
+For an invalid field, the default output for the jQuery Validation Plugin is an error message in a `<label>` element. This results in two `<label>` elements pointing to a single input field using the `for` attribute. While this is valid HTML, it has inconsistent support across screen readers.
+
```
staabm: maybe we should mention a example on how this "bad html" looks like

KatieMFritz:
This is great!

staabm:
Sorry for having this not merged earlier.

We can improve it further but lets have it in the readme for now.


Thx for the contribution and sorry for the overlong wait periode

PR #2154, opened 2018-03-20
Fixed bug for Html Editor(summernote)

vmtco:


Arkni:
Hi @vmtco,

I think we have this fixed in master, but not released yet. Could you please test with the version in master and see if the issue persist. You can grab the dist file from this gist: https://gist.github.com/Arkni/495730a47b52b570b150d4375bfbb7c7

Thanks!

markussc:
@Arkni : This fix is not contained in your gist. I also tested and it does not work. With the bugfix by @vmtco, everything works fine. Please merge and release.

Arkni:
Hi @markussc,

Thanks for the feedback. I'll try to analyse the issue more and will offer feedback as soon as I get some free time.

KoalaBear84:
@Arkni 

Ouch, something that could already be fixed for years, but isn't. I get it, free time is hard for me too. 😅

It would fix it for Trumbowyg which we just tested, and also CKEditor, and probably a lot more alternative WYSIWIG editors.

KoalaBear84:
Thanks (in advance)! 😇

KoalaBear84:
It would be very nice if this could be published in a new release that will be on npm so it will be fixed for everyone. Thanks in advance.

bytestream:
@KoalaBear84 `npm i jquery-validation@1.19.4-pre`

PR #2072, opened 2017-09-13
Quick update for the 2 new danish pull-request...

BenjaminHoegh:
Quick update for the 2 new danish pull-request there adds Danish phone number and postal codes

Add translate for:
 - phoneDK
 - postalcodeDK

PR #2165, opened 2018-04-24
Danish Update

BenjaminHoegh:
GitHub has somehow made me lose the permissions to edit #2070 #2071 and #2072. So here is a new pull-request with the changes and all-in-one request (my old pc couldn't run GitHub sorry for that).

### Accepts:
+45 21-77-21-77
+45 21 77 21 77
+45 21772177
+4521772177
21 77 21 77
21-77-21-77
2177 2177
2177-2177
21772177

Have changed the regex to use the non-capturing group as requested.

## Postal codes

Have also changed the postal codes to accept from the Danish Kingdom instead of Danish and Greenland only. This means that the Faroe Islands is now included.

### DIFF
--- a/src/additional/phoneDK.js
+++ b/src/additional/phoneDK.js
@@ -1,0 +1,27 @@
+/**
+ * Matches DK phone number format
+ *
+ * allows '-' or ' ' as a separator and allows parens around area code
+ * some people may want to put a '+45' in front of their number
+ *
+ * +45 21-77-21-77
+ * +45 21 77 21 77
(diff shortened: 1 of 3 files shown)


BenjaminHoegh:
Have no idea about why its keep failing the test

Arkni:
Hi @BenjaminHoegh 

Thanks for your contribution! and sorry for taking so long to repond to this PR. 
I'll triage and take a look at it as soon as possible!

PR #2174, opened 2018-06-01
Remote validation needs to handle failed requests

das-peter:
#### Description
The current implementation of the ajax handling in the remote validation seems to care about successful requests.
This might lead to not properly validated data.
We should use the `complete` callback instead `success` - this should be doable with fairly minimal changes to the current function.


Bobinho:
I know that this PR is quite old, but is there a chance to pull this changes? Seems like my team almost reintroduced the wheel because of need of such fix.

Issue #2127, opened 2018-01-15
Install StaleBot to manage stale issues?


jzaefferer:
This bot looks pretty useful: https://github.com/apps/stale

Found it here: https://remysharp.com/2018/01/10/open-source-with-a-cap-in-hand

We could enable that for this repo and let it shut down a bunch of old issues...

staabm:
Sounds great

Arkni:
We can also install [`no-response`](https://github.com/probot/no-response) bot to automatically close Issues where the author hasn't responded to a request for more information.

staabm:
I dont have experience with any of those bots, so please use whatever you guys think fits best

Arkni:
Will try to configure the mentioned bots soon.

staabm:
Stale bot installed with #2176

PR #2176, opened 2018-06-04
Add support for Stale bot

Arkni:
I will install the bots on this repo after we agree on the message of the comments.

Ref #2127

/cc @staabm

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,24 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - bug
+  - "help wanted"
+  - "MERGE ME"
+  - "NEEDS REVIEW"
+# Ignore issues in a milestone
+exemptMilestones: true
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue/proposal has been automatically marked as idle and stale because it hasn't
+  had any recent activity. It will be automtically closed if no further activity
+  occurs. If you think this is wrong, or the problem still persists, just pop
+  a reply in the comments and one of the maintainers will (try!) to follow up.
+
+  Thank you for contributing :)
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false


.github/stale.yml
```diff
@@ -0,0 +1,20 @@
...
+  - bug
+  - "MERGE ME"
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as idle and stale because it hasn't
+  had any recent activity. It will be automtically closed if no further activity
+  occurs. If you think this is wrong, or the problem still persists, just pop
+  a reply in the comments and @staabm or @Arkni will (try!) to follow up.
```
staabm: please dont mention us... the bot will spam us with notifications ;-)

.github/stale.yml
```diff
@@ -0,0 +1,20 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - bug
+  - "MERGE ME"
```
staabm: should we add "help wanted" here?

staabm: and also "NEEDS REVIEW"?

.github/stale.yml
Arkni: Haha, I forgot why I put them in there.

I will replace them with `one of the maintainers`

.github/stale.yml
Arkni: You're right. I should add those too.

Arkni:
@staabm Updated per above suggestions

Arkni:
As far as I know, none of the two bots support a rollback action.

The `no-response` bot only support 3 options, which we already use. As for the `Stale` bot, there are still some usefull options to use. For example, `exemptMilestones` that can be used to ignore issues or pull requests that are part of a milestone. We can also use `only` option to limit the to issues only or pull requests (by default, it manage the two).

For more options, see https://github.com/probot/stale#usage

Arkni:
Also, forgot to mention that the `no-response` bot has a annoying issue when the original poster closed the issue, see https://github.com/probot/no-response/issues/16

staabm:
maybe we should start with the "stale" bot only and see how things goes.

the other bot config file could be stored in a separate PR for later merge.

Arkni:
Sounds good to me. I will update the PR later today.

Arkni:
Update per above suggestion. I think we are good to go.

Arkni:
I broke the PR into two:
- This for `Stale` bot
- #2178 for `No-response` bot

staabm:
ok cool, thank you.

Arkni:
I have sent an installation request to `jquery-validation`. As you're the one who created this org, only you have the right to accept the request.

jzaefferer:
Should the bot apply to all repos, or only this one?

staabm:
we don't have "Issues"-tab activated in the other reposes

jzaefferer:
I guess you already approved it? I'm not sure based on the email notification and Apps UI...

staabm:
yep, I did.

PR #2183, opened 2018-06-19
Core: simplify destroy method

Arkni:


staabm:
Not sure we can do this change, as people might rely on this classnames for css and/or js.

A github code search could give an answer to this question...

Arkni:
I did a research and all I can find is our own source code. I even check the source code of [`jquery-validation-unobtrusive`](https://github.com/aspnet/jquery-validation-unobtrusive) which is one of the famous plugins that depend on `jQuery Validation`. I also checked [`wet-boew/wet-boew`](https://github.com/wet-boew/wet-boew) which heavily use this plugin for validation.

But I guess, it would be best suitable for v2.

/cc @RobJohnston: wdyt about this?

Arkni:
Thanks for the input @RobJohnston 
@staabm I'll play it safe regarding this and schedule it for v2

PR #2186, opened 2018-06-27
All: Increase number of days of inactivity of stale issues to 60 days

Arkni:
Hi @staabm,

This will help us more to not miss some important issues. 30 days is good, but sometimes I want to address an issue and a month pass quickly.

Also, do you think we need to add some labels (e.g `question`) to the list of excluded labels?


staabm:
> Also, do you think we need to add some labels (e.g question) to the list of excluded labels?

yeah, I guess this is a good idea.

PR #2207, opened 2018-09-03
mobile number for Russia Federation

jehadja:
#### Description

Thank you!

### DIFF
--- a/src/additional/mobileRU.js
+++ b/src/additional/mobileRU.js
@@ -1,4 +1,4 @@
 $.validator.addMethod( "mobileRU", function( phone_number, element ) {
 var	ruPhone_number = phone_number.replace( /\(|\)|\s+|-/g, "" );
-	return this.optional( element ) || ruPhone_number.length > 9 && /^((\+7|7|8)+([0-9]){10})$/.test( ruPhone_number )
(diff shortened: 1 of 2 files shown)


Arkni:
Thanks for your contribution! I'll triage and take a look at it as soon as possible!

src/additional/mobileRU.js
```diff
@@ -0,0 +1,6 @@
+/*  mobile number for Russia Fedaration */
+$.validator.addMethod( "mobileRU", function( phone_number, element ) {
+	phone_number = phone_number.replace( /\(|\)|\s+|-/g, "" );
```
Arkni: Please create a new variable `phoneNumber` or something else and assign the result to it.

src/additional/mobileRU.js
```diff
@@ -0,0 +1,6 @@
...
+	return this.optional( element ) || phone_number.length > 9 &&
+		phone_number.match( /^((\+7|7|8)+([0-9]){10})$/ );
```
Arkni: Please use `regex.test` instead of `string.match`. We only need to know whether the string matches the regex or not.
`phone_number.match( /^((\+7|7|8)+([0-9]){10})$/ )` => `/^((\+7|7|8)+([0-9]){10})$/.test( phone_number )`

PR #2213, opened 2018-09-13
Fix: bug with summernote code

kurybr:
#### Description

I made a filter that identifies if it has the class ".note-editable", if it exists it searches for the "parent" that has the textarea and then it changes the element by the correct element!

Good Job ! 

Thank you!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -268,6 +268,7 @@ $.extend( $.validator, {
 		ignore: ":hidden",
 		ignoreTitle: false,
 		onfocusin: function( element ) {
+			element = this.isSummernote(element); // To Fix Bug with Summernote Plugin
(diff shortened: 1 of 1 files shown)


Arkni:
Hi @kurybr,

Thanks a lot for the PR, but we can't accept framework/editors specific hacks into core. Also, this issue is not something related to summernote only, it's a known issue from our side but unfortunately we don't have enough time to look into it.

As a quick solution for now, see my comments https://github.com/jquery-validation/jquery-validation/issues/1875#issuecomment-272667183 and https://github.com/jquery-validation/jquery-validation/issues/1875#issuecomment-314919732.

Thanks again for the issue and the PR.

PR #2216, opened 2018-09-24
ABARoutingNumber: Add ABA Routing Number Validation

coryasilva:
#### Description

This is used in the United States to process payments, deposits, or transfers using the Automated Clearing House (ACH) or Fedwire systems.  A very common use case would be to validate a form for an ACH bill payment.

https://en.wikipedia.org/wiki/ABA_routing_transit_number


Arkni:
Thanks for your contribution! I'll triage and take a look at it as soon as possible!

staabm:
The filenames should also use lower letters.

Arkni:
Good catch @staabm.

@coryasilva could you please rename the source and test files to be the same as the method name? After that, we are good to go.

Thanks!

coryasilva:
lol, damn windows box.  The company finally gave me a macbook, give me a couple more days to migrate my dev environment and I will make sure the file renames stick.

coryasilva:
Okay I just renamed the files in github.  I hope that worked.

PR #2230, opened 2018-10-25
component: Chrome autofill bugfix

szmagli:
resolve jquery-validation/jquery-validation#2228

#### Description

Thank you!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -58,7 +58,7 @@ $.extend( $.fn, {
 					//   - A user defined a `submitHandler`
 					//   - There was a pending request due to `remote` method and `stopRequest()`
 					//     was called to submit the form in case it's valid
-					if ( validator.submitButton && ( validator.settings.submitHandler || validator.formSubmitted ) ) {
+					if ( validator.submitButton && ( validator.settings.submitHandler || validator.formSubmitted ) && !validator.settings.disableSubmitSec ) {
 						hidden = $( "<input type='hidden'/>" )
 							.attr( "name", validator.submitButton.name )
 							.val( $( validator.submitButton ).val() )
@@ -261,6 +261,7 @@ $.extend( $.validator, {
 		validClass: "valid",
 		errorElement: "label",
 		focusCleanup: false,
+		disableSubmitSec: false,
 		focusInvalid: true,
 		errorContainer: $( [] ),
 		errorLabelContainer: $( [] ),


staabm:
Could you elaborate a bit why this change fixes your problem?

The newly added config key is rather cryptic named, I dont know what it does...?

Why do we need this new config flag after alll?

szmagli:
Any form element (button, input etc.) inserted by javascript trigger chrome autofill.
I should rename config option to disableSubmitButton.
This should be a new flag becouse this options is required for 'remote' function.

src/core.js
```diff
@@ -58,7 +58,7 @@ $.extend( $.fn, {
 					//   - A user defined a `submitHandler`
 					//   - There was a pending request due to `remote` method and `stopRequest()`
 					//     was called to submit the form in case it's valid
-					if ( validator.submitButton && ( validator.settings.submitHandler || validator.formSubmitted ) && !validator.settings.disableSubmitSec ) {
+					if ( validator.submitButton && ( validator.settings.submitHandler || validator.formSubmitted ) && !validator.settings.disableHiddenInput ) {
```
staabm: comment above is now outdated.

staabm:
as I read the comment in the source, I am wondering, why you use a form without a explicit submit button?

src/core.js
szmagli: Comment changed

szmagli:
Any variation of form doesn't work (even from example files). 

If i doesn't disable this option by `disableHiddenInput` this hidden input is created.
I always use submitHandler to process data after validation and `validator.submitButton` always is true

staabm:
hmm not sure this solution is a good fit. @Arkni  any opinions/ideas on thiw?

szmagli:
I'm also not sure about this solution. When someone change password and doesn't confirm this to chrome when he try to login injecting any button, input etc. restore form fields. I need find way to fix this, not skip.

Arkni:
Hi @szmagli,

I can't test it right now, but could setting `autocomplete` to `"off"` (or to a random string) on the hidden input stop the browser from auto filling the other fields?

szmagli:
@Arkni i tried but onload aoutocomplete should be on, before i call validation i try to change to off, password field to autocolplete="new-password" but chrome ignore that.

src/core.js
```diff
@@ -54,11 +54,11 @@ $.extend( $.fn, {
...
 
 					// Insert a hidden input as a replacement for the missing submit button
-					// The hidden input is inserted in two cases:
+					// The hidden input is inserted when user doesn't define `disableHiddenInput` as true and one of these two cases are fulfilled:
 					//   - A user defined a `submitHandler`
 					//   - There was a pending request due to `remote` method and `stopRequest()`
 					//     was called to submit the form in case it's valid
-					if ( validator.submitButton && ( validator.settings.submitHandler || validator.formSubmitted ) ) {
+					if ( validator.submitButton && ( validator.settings.submitHandler || validator.formSubmitted ) && !validator.settings.disableHiddenInput ) {
 						hidden = $( "<input type='hidden'/>" )
```
staabm: Arknis idea was to add autocomplete off to this input which is beeing injected

szmagli: it doesn't work neither

PR #2234, opened 2018-10-31
 Create cnhBR.js

joaoissamu:


src/additional/cnhBR.js
```diff
@@ -0,0 +1,49 @@
...
+  if ( value.length !== 11 ) {
+    return false;
+  }
+
+  var sum = 0, dsc = 0, firstChar,
+		firstCN, secondCN, i, j, v;
+
+  firstChar = value.charAt( 0 );
+
+  if ( firstChar.repeat( 11 ) === value ) {
```
Arkni: `string.repeat()` is an ES6 feature and we still support IE. So can you please replace it with:
```suggestion
  if ( new Array(12).join( firstChar ) === value ) {
```

PR #2235, opened 2018-11-04
Core: Fix contenteditable detection's regression introduced in #2142

### LINKED ISSUE #2211: regression after fix for bug #2141 (jquery-validation 1.18.0 doesn't work anymore with IE11)
Hi,
I've tried latest 1.18.0 release, sadly the patch #2142 for bug #2141 introduced a bad regression on (at least) IE11, because validation doesn't work anymore!

The culprit is following code path in jquery.validate.js:

			if ( element.isContentEditable ) {
				val = $element.text();
			} else {
				val = $element.val();
			}

When you try my code example at https://liveweave.com/N6xRPw using latest Chrome or Firefox, validation works correctly because element.isContentEditable is false.
But when you use IE11, element.isContentEditable seems to be always true and jquery-validation "sees" a blank field because the variable "val" is always set to "". On the other side, element.hasAttribute( "contenteditable" ) correctly returns false on IE11.

Reverting the patch #2142 solves the problem for me, and everything seems to work again like it did in 1.17.0

Please could you have a look at this problem and confirm me that you can reproduce it on an other IE11 installation?
It would be then great to have a 1.18.1 release, because it's a very nasty regression ;)

Thank you!

### LINKED ISSUE #2214: update from 1.17 to 1.18 breaks in IE 11
### Subject of the issue

been using this on my site for years, recently upgraded 1.17 to 1.18 and everything still works fine except when visiting my site on IE 11.  A basic example is that on my login page, i have a username/password that are both required.  When i submit I get a client side error telling me that the username is required, even though the username field is filled in.

### Your environment
* version of `jquery-validate`:  1.18
* which browser and its version:  IE 11

### Expected behaviour

if a value is in the text field, it should be recognized as being filled in
### Actual behaviour

input box value is ignored.  can't move on.

### LINKED ISSUE #2217: Issue with v1.18.0, IE and Required
### Summary of Issue
Since upgrading to v1.18.0, text boxes trigger a required validation error even though there is text in the text box. It does not happen in chrome. Rolling back to v1.17.0 fixed the issue.

### Your environment
* version of `jquery-validate`: v1.18.0
* which browser and its version: Internet Explorer v11

### Steps to reproduce
Simple MVC Core 2.1 application with a text box created using Html.TextBoxFor.  The input element is generated with data-val="true".  Html.ValidationFor used to generate the validation error.

### Expected behaviour
Would expect no validation error when submit clicked. Normally then when typing in the field, the validation error cleared.

### Actual behaviour
Validation error persists and cannot submit form.

Extra debugging revealed that when jquery-validate checked if the element was editable (isContentEditable), it then either did val = $element.text() or val = $element.val(). In the case of my text box, the isContentEditable flag was true, get $element.text() returned "".  Hence the required validation failed. $element.val() return the actual contents of the textbox, but this was not used.

### LINKED ISSUE #2225: 1.18.0 fails in Internet Explorer 11
### Subject of the issue
Validation fails in Internet Explorer 11

### Your environment
jQuery Validate v1.18.0
Internet Explorer 11

### Steps to reproduce
```
<form id="RegistrationForm">
<input type="text" id="FirstName" name="FirstName"/>
<input type="submit" value="Submit"/>
</form>

<script type="text/javascript">

    jQuery('#RegistrationForm').validate({
        ignore: ".ignore",
        rules: {
            FirstName: {
                required: true
            }
        }
    });

</script>
```

### Expected behaviour
Not valid status should be removed from "required" fields after content is entered.

### Actual behaviour
Fields stay in the not valid status even after content is entered. The form cannot be submitted.

### LINKED ISSUE #2233: Required validation failed on IE 11
**After update the version to 1.18, I found the required validation is not work on IE 11.**

I did some debug and found `isContentEditable` always `true` on IE11 

```
<input type="text" id="userName" name="userName" class="form-control" data-val="true" data-val-required="Please input your name">
```
### Debug info
image

the $element has value
image

**I found 'isContentEditable' is from this commit:**
https://github.com/jquery-validation/jquery-validation/commit/7c22d68a4b906498f7d1b76a5751135f0d8266a6

Arkni:
Fixes #2211
Fixes #2214
Fixes #2217
Fixes #2225
Fixes #2233

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -125,6 +125,7 @@ $.extend( $.fn, {
 	// https://jqueryvalidation.org/rules/
 	rules: function( command, argument ) {
 		var element = this[ 0 ],
+			isContentEditable = typeof this.attr( "contenteditable" ) !== "undefined" && this.attr( "contenteditable" ) !== "false",
 			settings, staticRules, existingRules, data, param, filtered;
 
 		// If nothing is selected, return empty object; can't chain anyway
@@ -132,7 +133,7 @@ $.extend( $.fn, {
 			return;
 		}
 
-		if ( !element.form && element.isContentEditable ) {
+		if ( !element.form && isContentEditable ) {
 			element.form = this.closest( "form" )[ 0 ];
 			element.name = this.attr( "name" );
 		}
@@ -393,9 +394,10 @@ $.extend( $.validator, {
 			} );
 
 			function delegate( event ) {
+				var isContentEditable = typeof $( this ).attr( "contenteditable" ) !== "undefined" && $( this ).attr( "contenteditable" ) !== "false";
 
 				// Set form expando on contenteditable
-				if ( !this.form && this.isContentEditable ) {
+				if ( !this.form && isContentEditable ) {
 					this.form = $( this ).closest( "form" )[ 0 ];
 					this.name = $( this ).attr( "name" );
 				}
@@ -629,12 +631,14 @@ $.extend( $.validator, {
 			.not( this.settings.ignore )
 			.filter( function() {
 				var name = this.name || $( this ).attr( "name" ); // For contenteditable
+				var isContentEditable = typeof $( this ).attr( "contenteditable" ) !== "undefined" && $( this ).attr( "contenteditable" ) !== "false";
+
 				if ( !name && validator.settings.debug && window.console ) {
 					console.error( "%o has no name assigned", this );
 				}
 
 				// Set form expando on contenteditable
-				if ( this.isContentEditable ) {
+				if ( isContentEditable ) {
 					this.form = $( this ).closest( "form" )[ 0 ];
 					this.name = name;
 				}
@@ -689,6 +693,7 @@ $.extend( $.validator, {
 		elementValue: function( element ) {
 			var $element = $( element ),
 				type = element.type,
+				isContentEditable = typeof $element.attr( "contenteditable" ) !== "undefined" && $element.attr( "contenteditable" ) !== "false",
 				val, idx;
 
 			if ( type === "radio" || type === "checkbox" ) {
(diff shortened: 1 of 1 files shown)


staabm:
Cool. Can we get this unit tested somehow?

Arkni:
I think we already have some tests for contenteditable elements. One thing I can suggest in order to avoid regressions regarding IE, is for us to apply for a Saucelabs/BrowserStack open source access to use it for testing the plugin in supported IE versions and also on Safari (if it's allowed in the free plan).

staabm:
Oh I see. Not sure it’s worthwhile for IE. In case you are condfident with the patch please go ahead and merge it.
 If not we can wait for som of the reporters to confirm the fix.

Thank you

Arkni:
It's best to wait for feedback from the reporters. I will provided the build file for them and link to it in #2211.

PrinceCharles23:
judging from the code change this seems to do the trick. since i get the package via npm i have no clue how to test the change directly in our project...?

staabm:
see https://github.com/jquery-validation/jquery-validation/issues/2211#issuecomment-435686142 on how to get the fixed version

PrinceCharles23:
> see [#2211 (comment)](https://github.com/jquery-validation/jquery-validation/issues/2211#issuecomment-435686142) on how to get the fixed version

this just links to the fixed file. npm would overwrite this during next build

Arkni:
Based on the feedback from the comunity, I am going to merge this.

Arkni:
@staabm could you please release this fix in a minor version (as we also have some new validation rules added)?

Thanks :)

staabm:
just published 1.19.0 with a fix for this issue.

sorry the any inconvenience this bug has caused for you guys.

PR #2236, opened 2018-11-09
Additional (CNPJBR and CPFBR): Fix not to fail when field is not required

jcspader:
Additional: CNPJBR and CPFBR

 Fix not to fail when field is not required


Arkni:
Thanks for your contribution :)

PR #2237, opened 2018-11-21
Updating bug for fields with no Name nor Id

seltix5:
When using some WYSIWYG like summernote, by default, contenteditable area has not Id nor Name, in this cases the script generates error in "escapeCssMeta()" when trying to do replace on undefined because the string arg is the result of "idOrName()" on line #1009.

PS : I dont know how the tests work nor how to create one... If you can send me some info about it I could do it.

DEMO ERROR : http://jsfiddle.net/3bpnw7uo/8/


bytestream:
@staabm This can be closed as it duplicates https://github.com/jquery-validation/jquery-validation/pull/2154 which has now been merged?

staabm:
Good point, thx

PR #2243, opened 2018-12-04
Core: change focus() to trigger("focus")

sassquad:
#### Description
This PR changes the focus() to trigger("focus"), so that it is not flagged up in non-live environments using the jQuery Migrate 3.x plugin.

https://github.com/jquery-validation/jquery-validation/issues/2238


Arkni:
Thanks for your contribution! I'll take a look at it as soon as possible!

sassquad:
No problem @Arkni 👍

PR #2247, opened 2018-12-13
elementValue fix for editable content

bogatykh:
<input type="text" is editable, so .val() should be used instead of text()


staabm:
Which problem does this change solve? How to reproduce?

Arkni:
Hi @bogatykh,

If you're trying to fix the same issue as #2211, it was already fixed and released as part of version `1.19.0`. Otherwise, please provide more information about your issue.

Thanks!

PR #2248, opened 2018-12-16
set jQuery as a peer dependency

aigdonia:
#### Description

As fix for #1858
Thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
 		"dist/jquery.validate.min.js"
 	],
 	"main": "dist/jquery.validate.js",
-	"dependencies": {
+	"peerDependencies": {
 		"jquery": "^1.7 || ^2.0 || ^3.1"
 	},
 	"devDependencies": {


Arkni:
I'm +1 to this change. But, is it ok to release this in a minor/patch version instead of a major one?

aigdonia:
any update to decision here?

Arkni:
Hi @aigdonia,

Sorry for taking so long to respond, the last month (and still have limited time) was crazy for me. As I said before, I'm +1 to the idea of removing jQuery from the dependencies and moving it to peer dependencies. The only thing I lack information on is should we release this as part of a patch/minor version or a major one?

If you have any information/recommendation about that, please feel free to provide them.

Thanks a lot for working on this and thanks for your patience :)

aigdonia:
as per the definition of [semver.org](https://semver.org/)
a MAJOR release should make incompatible API changes, which is not the case here,
this change won't affect the current installs of the package as well,
the only effect I see it when new `npm install` is introduced without `jquery` being introduced already in the project, there will be a warning to install it.

so I suggest this to be a minor/patch release.

Arkni:
A minor release it is then.

Thanks @aigdonia for the information :) I'm going to merge this proposal in a moment.

aigdonia:
Thanks for the merge

ericfowler303:
Would it be possible to publish a beta build or public build of this to bring in via npm? I'm currently having issues with this using webpack due since this commit isn't in a built package.

tommyip:
Until a new build is published, simply add the following to `package.json` to fix the problem:
```
"resolutions": {
    "jquery-validation/jquery": "3.4"
}
```
(Thanks @marcwieland95 for coming up with the fix in #2272)

pam81:
I tried this solution but it's still using the jquery from the inside node_module folder. 
The only way i found a solution is to removed this folder after doing yarn install.

staabm:
I just published jquery-validateion 1.19.1 to npm.

most important change is that jquery is now defined as a peer dependency, so we dont get into your way. see https://github.com/jquery-validation/jquery-validation/pull/2248

sorry this took us so long. jq-validation is not top priority atm

PR #2261, opened 2019-02-12
Core: Instead of fetching only child of form, fetch all elements & filter records.

garima2094:
#### Description
**The plugin does not validate elements that are outside of the `<form>` but have the `form` attribute on them.**

This issue occurred because in the plugin we only find input fields which are the child of form but we did not get all fields which are outside the form but associated with the form.

after analyzing the code I found the solution, at line no. 646 Did small change and it works for me.

`// Select all valid inputs inside the form (no submit or reset buttons) return $( this.currentForm ) .find( "input, select, textarea, [contenteditable]" )`

Changed above code to this
`// Select all valid inputs inside the form (no submit or reset buttons)
return $.merge( $( this.currentForm ).find( "input, select, textarea, [contenteditable]" ),
            $( this.currentForm.elements ).filter( "input, select, textarea, [contenteditable]" ) )
Thank you!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -625,8 +625,8 @@ $.extend( $.validator, {
 				rulesCache = {};
 
 			// Select all valid inputs inside the form (no submit or reset buttons)
-			return $( this.currentForm )
-			.find( "input, select, textarea, [contenteditable]" )
+			return $( this.currentForm ).fields()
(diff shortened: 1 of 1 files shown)


Nuranto:
Thanks for the fix ! Hope it will be merge soon :)

PR #2289, opened 2019-07-09
Core: Add check undefined or null for escapeCssMeta

henryhuang:
#### Description
For some text editors (https://github.com/Alex-D/Trumbowyg etc.) integrating, their editor element will not set name attribute or id attribute, then core.js will throws 

```
jquery.validate.min.js:4 Uncaught TypeError: Cannot read property 'replace' of undefined
    at a.validator.escapeCssMeta (jquery.validate.min.js:4)
    at a.validator.errorsFor (jquery.validate.min.js:4)
    at a.validator.prepareElement (jquery.validate.min.js:4)
    at a.validator.element (jquery.validate.min.js:4)
    at a.validator.onfocusout (jquery.validate.min.js:4)
    at HTMLDivElement.b (jquery.validate.min.js:4)
    at HTMLFormElement.dispatch (jquery.min.js:2)
    at HTMLFormElement.y.handle (jquery.min.js:2)
    at Object.trigger (jquery.min.js:2)
    at Object.simulate (jquery.min.js:2)
```
This pull request fix this issue.

Thank you!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -950,7 +950,9 @@ $.extend( $.validator, {
 		// meta-characters that should be escaped in order to be used with JQuery
 		// as a literal part of a name/id or any selector.
 		escapeCssMeta: function( string ) {
-			return string.replace( /([\\!"#$%&'()*+,./:;<=>?@\[\]^`{|}~])/g, "\\$1" );
+			if (string) {
+				return string.replace( /([\\!"#$%&'()*+,./:;<=>?@\[\]^`{|}~])/g, "\\$1" );
+			}
 		},
 
 		idOrName: function( element ) {


staabm:
having a name attribut is a fundamental requirement for this lib to work.

this will not be the only issue you will face I guess

henryhuang:
@staabm after I add a name attribute to the tag which generate by third part editor lib, it was work. For example, my code is 

```
<textarea id="content" name="content"></textarea>
```

but the editor lib will generate

```
<div class="editor"></div>
<textarea id="content" name="content" style="display:none"></textarea>
```

I think your lib should get textarea which id is content, but you get div which class is editor, then this error occurs.

gbjbaanb:
@henryhuang example should probably also have the contenteditable="true" flag is set on the parent div to the textarea. the library uses a div to edit, but stores the data in the textarea for validation.

The system is not trying to validate the div, but as it has the contentEditable flag, validation is being performed against it regardless.

I think the best system would be to not perform the validation at all if the element being checked does not have a name (or ID), as its apparently a requirement. This should probably be part of the function delegate that performs the isContentEditable check.
```

<div class="trumbowyg-editor" dir="ltr" style="height: 110.583px;" contenteditable="true">
  <p>fgdfgfd<br></p>
</div>
<textarea id="taSP" rows="5" class="form-control trumbowyg-textarea" data-val="true" name="SpecialOffers"></textarea>
```

staabm:
as stated, having a name is a fundamental requirement, we cannot change

> 
> 
> having a name attribut is a fundamental requirement for this lib to work.
> 
> this will not be the only issue you will face I guess

gbjbaanb:
Its not a fundamental requirement if you're not trying to validate the field!

The code tries to validate fields it should not. fields that do not have a name because they are not being validated. But the code still tries to run its validation. This is a bug.

Now the pragmatic option is to ignore those fields that do not have the attributes that are required for validation - such as a name. So if a field has no name, instead of blindly throwing an exception, it should more gracefully handle this error state. 

That's good programming practice - always validate your inputs and do not attempt to process anything that fails to meet the minimum requirements. You'd surely agree that's a acceptable thing to do, right 😉

tolbxela:
You should validate inputs in the functions! This belongs to quality programming code.

PR #2330, opened 2020-05-14
Update core.js

robuuust:
fix for null reference exception

#### Description

Thank you!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1032,7 +1032,7 @@ $.extend( $.validator, {
 		// meta-characters that should be escaped in order to be used with JQuery
 		// as a literal part of a name/id or any selector.
 		escapeCssMeta: function( string ) {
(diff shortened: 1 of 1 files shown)


staabm:
thx for the PR. please describe your use case and show us why this change is required and how to reproduce your problem

robuuust:
When using summernote editor (https://www.npmjs.com/package/summernote) we get an error on this line because string happens to be undefined. The escapeCssMeta is receiving an undefined input instead of a string value. This causes the entire page to crash resulting in a broken editor.
This extra null check solves this issue.

staabm:
in my experience this kind of change only hides actual errors at the caller side.

If we dont have evidence this is usefull for a lot of upstream projects, I dont want to merge this.

staabm:
likely you run into this problem, because the element beeing validated misses a `name` attribute

PR #2335, opened 2020-06-23
Replaced deprecated jQuery functions

bytestream:
#### Description

Replaces functions deprecated in jQuery 3 as per https://api.jquery.com/category/deprecated/

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1313,7 +1313,7 @@ $.extend( $.validator, {
 
 		// Evaluate parameters
 		$.each( rules, function( rule, parameter ) {
-			rules[ rule ] = $.isFunction( parameter ) && rule !== "normalizer" ? parameter( element ) : parameter;
+			rules[ rule ] = typeof parameter === "function" && rule !== "normalizer" ? parameter( element ) : parameter;
 		} );
 
 		// Clean number parameters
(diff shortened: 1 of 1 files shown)


src/core.js
```diff
@@ -205,6 +205,13 @@ var trim = function( str ) {
 	return str.replace( /^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, "" );
 };
 
+// $.isArray is deprecated.
+var isArray = function( arg ) {
+
+	// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Polyfill
+	return Object.prototype.toString.call( arg ) === "[object Array]";
+};
+
```
Arkni, line 214: Why not using `Array.isArray()` instead of polyfilling it? It is supported by IE9+ so it's fine to use it instead.

I don't remember which versions we support, but IE9 & IE10 are already dead.

bytestream: I wasn't sure what browser versions this lib supports. If it matches jquery then I guess our answer is here: https://jquery.com/browser-support/

Happy to remove the polyfill if you are

staabm: atm we support the following jquery versions:

https://github.com/jquery-validation/jquery-validation/blob/af445b30fc6bbe7431fd8677ddad1008b866bdda/package.json#L39

I also dont like supporting IE9/10, but I guess we have to right now. Also, this is just a few lines of code, so no problem either.

please add a comment mentioning we need this polyfill for IE9/10 support and we can drop it, after rasing min IE to 11+

bytestream: The polyfill link says Array.isArray is compatible with IE 9+

staabm: > The polyfill link says Array.isArray is compatible with IE 9+

jepp, just add a comment in the code that we need this thing for IE9/10 compat and we are good to go.

staabm: > The polyfill link says Array.isArray is compatible with IE 9+

ohh, then we can just just `Array.isArray` instead of the polyfill. awesome.

bytestream: Awesome. I've removed it :)

PR #2340, opened 2020-08-10
Fix validation for WYSIWYG editors

### LINKED ISSUE #2229: WYSIWYG plugins causes "TypeError: string is undefined"
### Subject of the issue
WYSIWYG plugins causes "TypeError: string is undefined"

### Your environment
* version of `jquery-validate`: 1.18.0
* which browser and its version: Chrome 70.0.3538.67 x64 and Firefox 62.0.3 x64

### Steps to reproduce
Using WYSIWYG (tested with Gijgo and TinyMCE) type anything on text field.
Click outside.

### Expected behaviour
Just no Errors

### Actual behaviour
Logging Error: TypeError: string is undefined
Line 1023

### Fix suggestion
Just force variable string to be string everytime on line 1023:
Replace:
> 1023 `return string.replace..`
To:
> 1023 `return (string+"").replace..`

bartmathijssen:
Fixes #2229

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1032,7 +1032,7 @@ $.extend( $.validator, {
 		// meta-characters that should be escaped in order to be used with JQuery
 		// as a literal part of a name/id or any selector.
 		escapeCssMeta: function( string ) {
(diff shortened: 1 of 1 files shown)


KoalaBear84:
Another PR just got merged which is like this one: #2154

PR #2345, opened 2020-09-05
docs: Fix simple typo, atteched -> attached

timgates42:
There is a small typo in test/test.js.

Should read `attached` rather than `atteched`.


staabm:
Thx

PR #2348, opened 2020-10-05
Additional: cifES

egprida:
Bug when codes beginning with lower case letter.

#### Description

Thank you!

### DIFF
--- a/src/additional/cifES.js
+++ b/src/additional/cifES.js
@@ -55,6 +55,8 @@ $.validator.addMethod( "cifES", function( value, element ) {
 		return true;
 	}
 
+	value = value.toUpperCase();
+
 	var cifRegEx = new RegExp( /^([ABCDEFGHJKLMNPQRSUVW])(\d{7})([0-9A-J])$/gi );
 	var letter  = value.substring( 0, 1 ), // [ T ]
 		number  = value.substring( 1, 8 ), // [ P ][ P ][ N ][ N ][ N ][ N ][ N ]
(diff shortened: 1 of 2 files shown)


RobJohnston:
I went to the Wikipedia link provided in the source code, and then to the referenced documents at the "Agencia Estatal Boletín Oficial del Estado".  Although all the examples show an upper-case letter, I saw nothing that said it was **required** to be upper-case.  

Full discretion, I'm functional in Spanish, and use it almost every day in some way, but I'm far from fluent.  I stand to be corrected.

jarey:
Hello @RobJohnston ,

As you said is not required that the NIF uses upper-case letters only.
The problem with the actual code (and wich this pull request solves) is that the current regexp only works with upper case letters.
If you see the detail of the PR, the change only applies upper-case transformation to the NIF letter in order to match the existent regexp and solve the validation problem that exists in case that is specified a NIF with lower-case letter.

Regards.

test/methods.js
```diff
@@ -1555,7 +1555,7 @@ QUnit.test( "cifES", function( assert ) {
 	assert.ok( method( "S5141387J" ), "CIF valid" );
 	assert.ok( method( "S9803881C" ), "CIF valid" );
 	assert.ok( !method( "S98038813" ), "CIF invalid: digit control must be a letter (C)" );
-	assert.ok( method( "s98038813" ), "CIF valid: lower case" );
+	assert.ok( !method( "s98038813" ), "CIF invalid: lower case but digit control must be a letter (C)" );
```
staabm, line 1558: why is this considered invalid now?

in your comment you said lower case chars are fine.
what das "but digit control must be a letter" mean?

src/additional/cifES.js
```diff
@@ -55,6 +55,8 @@ $.validator.addMethod( "cifES", function( value, element ) {
 		return true;
 	}
 
+	value = value.toUpperCase();
```
staabm, line 58:
```suggestion
    // turn the validation case-insensitive
	value = value.toUpperCase();
```

PR #2357, opened 2020-10-30
Additional: Add validation rules for Chilean RUT

jmcandia:
#### Description

Thank you!


sunny-shu:
You'd better add test cases in test/additional.

RobJohnston:
I noticed that there is more than just the Chilean RUT in this pull request.  There is also a new method for a Chilean phone number and a different version of "lettersonly", which can't possibly apply only to one country.  I think you should use this PR for only the RUT validation and put the other two validations in separate PRs.

And as stated earlier, add unit tests.

PR #2360, opened 2020-11-18
Core: fix validation for input type="date"

### LINKED ISSUE #2359: input type="date" fails validation when using locale specific methods from localizations/
https://jqueryvalidation.org/date-method/ says: _This method is deprecated and will be removed in version `2.0.0`.
Please don't use it, since it relies on the `Date` constructor, which behaves very differently across browsers and locales. Use `dateISO` instead or one of the locale specific methods (in localizations/ and additional-methods.js)._

So we're using `localization/methods_pt.js`, and observing that it causes validation of date inputs to always fail, because they are unconditionally validated using the deprecated method, and not its suggested replacements.

### Environment
* `jquery@3.5.1`
* `jquery-validation@1.19.2`
* `jquery-validation-unobtrusive@3.2.11`
* Chrome 86.0.4240.111

### Steps to reproduce
* https://jsfiddle.net/1ua63jx0/
* focus into and out of the date input: it gets a red border, indicating `class="input-validation-error"`
* if `localization/methods_pt.js` is *not* loaded, then the validation succeeds -- even though our actual date format conforms to `methods_pt`

This is because date input value is always in ISO format, regardless of the display format [(reference)](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date)

### Expected behaviour
* `normalizeAttributeRule` should match `date` input type with `dateISO` rule, instead of `date` rule.

tyomitch:
No tests here, because none are possible with methods from localization/ (quoting https://github.com/jquery-validation/jquery-validation/blob/master/test/methods.js#L186 : "need to figure out how to test localized methods")

Fixes #2359

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1214,7 +1214,7 @@ $.extend( $.validator, {
 
 			// Exception: the jquery validate 'range' method
 			// does not test for the html5 'range' type
-			rules[ method ] = true;
+			rules[ type === "date" ? "dateISO" : method ] = true;
 		}
 	},
 


micschk:
Would be nice if this could be merged (assuming it indeed fixes date inputs).

staabm:
Could you verify wether it fixes the mentioned problem and also add a unit test?

nenadvicentic:
Hi,

I tested this change on `de` culture and it seems to work fine.

Overwritten original `rules[method] = true` line with  `rules[type === "date" ? "dateISO" : method] = true;` and `input type=date` client side validation started to work.

staabm:
In case we get a unit test for the expected behavior which passes with thxis change, we could merge the PR

nenadvicentic:
> In case we get a unit test for the expected behavior which passes with thxis change, we could merge the PR

I am not familiar with jQuery source or tests, so not sure if I personally can help. My intuition tells me, unit tests where for `input type=date` jquery-validation method `$(element).rules()` returns `{ dateISO: true}` instead of current `{ date: true }` would be sufficeint.

@staabm Any tip where this test can be added/implemented?

nenadvicentic:
@tyomitch, @staabm Boys, there example of your unit-test:

```js
QUnit.test( "rules() - returns dateISO for input type=date", function( assert ) {
	$( "#rangesMinDateInvalid" ).validate();
	assert.deepEqual( $( "#minDateInvalid" ).rules(), { dateISO: true, min: "2012-12-21" } );
} );
```

It can be added to `/test/rules.js` file. I used existing `input type=date` element in `index.html` - `#minDateInvalid`, which also contains `min` rule.

@tyomitch , @staabm How should we proceed with this?

tyomitch:
@nenadvicentic done, updated the patch with your test case.

The scenario which *should* be tested is that a localized date input passes validation; but perhaps @staabm can accept this as a passable approximation.

nenadvicentic:
> The scenario which _should_ be tested is that a localized date input passes validation; but perhaps @staabm can accept this as a passable approximation.

By specification value property of `input type=date` is always formatted as ISO8601 date `yyyy-mm-dd` regardless of localized formatting displayed by browser. Details can be found in Mozilla developer network [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date).

So test that `dateISO` validation is picked for `<input type="date">` is enough.

@tyomitch , @staabm : more I am thinking about this scenario, I am wondering if there is a need for validation at all in this case, since browser ensures that, if `value` is there, it has to be in correct format.

tyomitch:
I think this is wrong, and the problem still persists.

PR #2362, opened 2020-11-19
Prevent removal of error labels from other forms.

aperez987:
#### Description
This prevents the removal of error labels that belong to another form [mostly a hidden form] placed on the current form.
These error labels do not belong to any [not ignored] element on the current form, so they should not be removed.

#### Test
https://jsfiddle.net/Lngbcrfq/10/

Leave empty all inputs.
Press `Check Name` . An error label appears next to `Name` input.
Press `Check Email` . An error label appears next to `Email` input.

Press `Check Name`  again. The error label from `Name` disappears!

#### Comments
We sometimes use hidden forms bound to inputs on a display form ([because nested forms are not a thing](https://html.spec.whatwg.org/multipage/forms.html#the-form-element)).
The bound inputs don't have names, so they are not validated on the display form.
But the validation of the display form removes the error labels from the hidden form but keeps the `errorClass` on the inputs.
The effect is that inputs have an error, but the user does not know why.

This commit filters from `errors` labels that don't belong to any element in `elements`.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -670,8 +670,16 @@ $.extend( $.validator, {
 		},
 
 		errors: function() {
+			var context = this;
+			var getElementsNames = function ( elements ) {
+				return elements.map( function () { return context.idOrName( this ) });
+			}.bind( this );
+
+			var names = getElementsNames( this.elements() ).toArray();
 			var errorClass = this.settings.errorClass.split( " " ).join( "." );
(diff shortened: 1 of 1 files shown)


aperez987:
On further testing, this approach is really not ideal.
Would be much better to not add the labels to `this.toHide` in the first place.

Could someone guide me a little on this one?

PR #2363, opened 2020-11-27
Localization: Add "pattern" translation for French

RobJohnston:
Adds the localization for "pattern", in partial support of wet-boew/wet-boew#8983.


RobJohnston:
Can this one be squash committed?  I thought I could do it easily via the web interface, at the end of the day on a Friday, with somebody talking to me.

staabm:
thx, looks good.

PR #2365, opened 2020-12-01
additional: add semantic version (SemVer 2.0.0) validation

RobJohnston:
Matches a valid SemVer 2.0.0 format for versioning software using the regex given at https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

Includes the same unit tests as written in the example and translations for French and Spanish.


staabm (approved):
lgtm

PR #2369, opened 2020-12-18
Core: wait for pendingRequests to finish before submitting form

bytestream:
#### JS Fiddle

https://jsfiddle.net/bytestream/a16ehn0L/18/

#### Description

When jquery-validation is used with a WYSIWYG editor there can be a delay in synchronisation between the `div[contenteditable]` and the underlaying `textarea`. This delay combined with remote validation leads to a state where the form is valid but does not submit.

The issue is hopefully fairly obvious when looking at `stopRequest`:
```
if ( valid && this.pendingRequest === 0 && this.formSubmitted && this.form() ) {
```

`this.form()` can cause new remote validation rules to fire. In the vast majority of cases, the input data is the same and so jquery-validation looks at the cached result and doesn't bother with another AJAX request. In this case however, the delay in synchronisation causes the request data to change so jquery-validation fires another AJAX request. The line noted above currently does not wait for any new remote validation checks to return before submitting the form.

Thank you!

@staabm does this make sense?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1108,7 +1108,7 @@ $.extend( $.validator, {
 			}
 			delete this.pending[ element.name ];
 			$( element ).removeClass( this.settings.pendingClass );
-			if ( valid && this.pendingRequest === 0 && this.formSubmitted && this.form() ) {
+			if ( valid && this.pendingRequest === 0 && this.formSubmitted && this.form() && this.pendingRequest === 0 ) {
 				$( this.currentForm ).submit();
 
 				// Remove the hidden input that was used as a replacement for the
--- a/test/test.js
+++ b/test/test.js
@@ -80,6 +80,15 @@ $.mockjax( {
 	}
 } );
 
+$.mockjax( {
+    url: "stop-request-submit-form.php",
+    response: function() {
+        $( "input[name='username']" ).val( "foobar" ); // Simulate input data changing before request finishes.
+        this.responseText = "true"; // Data is valid.
+    },
+    responseTime: 1
+} );
+
 // Asserts that there is a visible error with the given text for the specified element
 QUnit.assert.hasError = function( element, text, message ) {
 	var errors = $( element ).closest( "form" ).validate().errorsFor( element[ 0 ] ),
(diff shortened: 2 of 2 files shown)


bytestream:
I worked around this using:
```js
submitHandler: function (form) {
    // CodeMirror saves when form submit events are fired. We need to manually trigger a save
    // so that jquery-validation has the correct form data.
    $(form).find('.CodeMirror').each(function(i, el){
        el.CodeMirror.save();
    });

    // more stuff here
},
```

That said, I believe there is still merit to this PR as `this.form()` should not be allow the form to submit if there are pending requests. The chances of this happening are slim for the reasons mentioned in my first message.

bytestream:
@staabm what's necessary to see this merged and released?

staabm:
The PR implies a synchronous remote validation step, right?

bytestream:
The remote validation is still asynchronous. `this.form()` fires asynchronous remote validation if the input value changes from the cached value, and `stopRequest` runs after each of those requests finish.

staabm:
yeah.. I was more about that the newly added condition will check directly after `.form() ` was invoked, which looks like a race condition involved when the validation is async?

bytestream:
I don't see a race condition. Eventually the pending remote validation requests will be 0, unless the input values repeatedly change then it would loop endlessly.

staabm:
thx!

bytestream:
@staabm would you mind publishing `1.19.4-pre` to npm?
```
npm publish --tag next
```

bytestream:
@staabm ?

staabm:
I will have a look this afternoon..

staabm:
@bytestream are you willing to help, if I grant you permission todo it yourself?

bytestream:
@staabm sure, username is bytestream on npm too

staabm:
just invted you on npm @bytestream 

please see the necessary steps at https://github.com/jquery-validation/jquery-validation/blob/eb88df0e635b19703dc60bbd7728537cc7b13cc9/build/release.js#L1-L22

bytestream:
@staabm thanks. I need GitHub access for some of those things, but I think I'm OK without to publish a `-pre` release. https://github.com/jquery-validation/jquery-validation/pull/2422 also needs merging as it blocks running the release

PR #2000, opened 2017-04-25
Fixed validation for named array within form validation (works for name="xyz" and name="xyz[]") both

saurabhiam:
#### Description
If a form input has a name="xyz[]" then only first element with name="xyz[]" is validated and rest are ignored.
This fix adds the flexibility for validating the named array (name="xyz[]") or a single name attribute (name="xyz") with no extra effort

Thank you!

PR #2431, opened 2022-07-01
Core: Add capability to check array of inputs

ridea-co:
#### Description
Add capability to check array of inputs.
Basically a copy & paste code from #2000 with checkbox/radio check

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -454,7 +454,16 @@ $.extend( $.validator, {
 		checkForm: function() {
 			this.prepareForm();
 			for ( var i = 0, elements = ( this.currentElements = this.elements() ); elements[ i ]; i++ ) {
-				this.check( elements[ i ] );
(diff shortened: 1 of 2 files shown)


bytestream:
What's the checkbox/radio check for?

ridea-co:
The reason of filtering out radios and checkboxes is that when you have multiple radio buttons or checkboxes with same name, I believe user really expecting only one of then to be selected instead all of them have to be selected.  So for those 2 types of inputs, they fall back to the original program logic thus error message should display correctly.

ridea-co:
Hi, just wondering if everything look good to be merged in?

Seb33300 (changes requested):
Please fix the requested change

bytestream:
Closing in favor of #2498 - please comment on that PR if it doesn't work / match the behaviour of this PR.

PR #2435, opened 2022-07-15
Core: fix race condition in remote validation rules

### LINKED ISSUE #2434: Race condition in remote validation rules
## Your environment
* Version of `jquery-validate`: 1.19.5
* Browser name and version: Google Chrome 103.0.5060.114 / Firefox 102.0.1 / Microsoft Edge 103.0.1264.49

## Current behavior

When a field has both a non-remote and a remote validation rule, if the server is a bit slow, a race condition can happen. The validation result for the previous value of a field replaces the one for its current value.

More precisely, for this bug to occur, the "new" value of the field must be something that will be considered invalid by the non-remote rule (which means the new value will not trigger a new Ajax request). A simple example is using a "required" rule and emptying the field, as shown in this live demo.

## Expected behavior

The validation result for the latest value of the field should be displayed instead of the validation result for its previous value.

## Live demo

https://codepen.io/bidord/pen/RwMoOeM

Both fields in this example have a "required" rule and a "remote" rule. The remote rule for the first field takes 0.1s, the remote rule for the second field takes 2secs.

Clicking "test" will:
 - Fill both fields with some random value (and trigger validation)
 - Wait for 1sec
 - Empty both fields (and trigger validation again)

We expect both fields to be displayed as invalid with a "Field is required." error. However you should see that the field with a slow remote validation rule is incorrectly displayed as valid.

You can trigger this bug "manually" as well : just fill the second field with a single character then quickly remove it.

bidord:
Fixes #2434

#### Description

- Add a new jQuery global method `$.ajaxAbort( port )` to the jQuery Ajax extension that is provided by this library. This method allows to abort an ajax request without sending a new one.
- Call this method at the beginning of `check()` method.

### DIFF
--- a/src/ajax.js
+++ b/src/ajax.js
@@ -1,5 +1,6 @@
 // Ajax mode: abort
 // usage: $.ajax({ mode: "abort"[, port: "uniqueport"]});
+//        $.ajaxAbort( port );
 // if mode:"abort" is used, the previous request on that port (port can be undefined) is aborted via XMLHttpRequest.abort()
 
 var pendingRequests = {},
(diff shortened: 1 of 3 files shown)


bytestream (approved):
Thanks for the effort you put into writing the issue, reproducer and time on this PR.

PR #2444, opened 2022-10-06
Core: added validations greaterThan and lessThan

humzayunasnandla:
New validation methods added for number for greaterThan and lessThan checks. 

` i.e  we can make a check on obtain marks and total marks of students `


bytestream:
Thanks but these methods already exist https://github.com/jquery-validation/jquery-validation/tree/master/src/additional

PR #2445, opened 2022-10-12
Localization: small improvement to pt_BR

feikname:
#### Description
"É requerido" although correct is weird in this context (it's much more used in formal settings for documents known as requerimentos) and too much of a literal translation. This change aims to improve user-friendliness in this particular message.

**Popularity of one vs the other:**
image

**Demonstration of how the adjective requerido is much more widely used in judicial settings:**
image

(Jus brasil is one of the go-to websites as a judicial source and "cpc"/"codigo civil" both refer to brazil's civil laws)

Thank you!

### DIFF
--- a/src/localization/messages_pt_BR.js
+++ b/src/localization/messages_pt_BR.js
@@ -6,7 +6,7 @@
 $.extend( $.validator.messages, {
 
 	// Core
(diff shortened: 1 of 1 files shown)


bytestream:
Thanks @feikname

Issue #2450, opened 2022-10-26
greaterThan() not normalizing input


enboig:
## Your environment
* Version of `jquery-validate`: v1.19.5
* Browser name and version: Firefox

## Current behavior
I need to compare 2 datetime fields. They are localized so I have writen a normalizer to transform them to "YYYY-MM-DD hh:mm" form (including leading zeros).
When I compare them with greaterThan() it fails because second date is not normalized
## Expected behavior

If input field has a normalizer, use it before comparing.

https://jsbin.com/goruriroto/edit?html,js,console,output

It may also affect greaterThanEqual, lessThan and lessThanEqual

enboig:
It's the same as #1964

enboig:
After using a "normalization" it ended in:
* If no normalization is used, "10" < "5" => true
* If normalization is used to return a Float, method required fails because `(10).length == 0`
So I ended creating my own methods:
```js
$.validator.addMethod( "greaterThanNumber", function( value, element, param ) {
    var target = $( param );

    if ( this.settings.onfocusout && target.not( ".validate-greaterThan-blur" ).length ) {
        target.addClass( "validate-greaterThan-blur" ).on( "blur.validate-greaterThan", function() {
            $( element ).valid();
        } );
    }
	var val = target.val(), normalizer;
	if (rules = target.rules()) {
		if ( typeof rules.normalizer === "function" ) {
			normalizer = rules.normalizer;
		} else if (	typeof this.settings.normalizer === "function" ) {
			normalizer = this.settings.normalizer;
		}
		// If normalizer is defined, then call it to retreive the changed value instead
		// of using the real one.
		// Note that `this` in the normalizer is `element`.
		if ( normalizer ) {
			val = normalizer.call( element, val );

			// Delete the normalizer from rules to avoid treating it as a pre-defined method.
			delete rules.normalizer;
		}

	}
    return Number.parseFloat(value) > Number.parseFloat(val);
}, "Please enter a greater value." );

$.validator.addMethod( "greaterThanEqualNumber", function( value, element, param ) {
    var target = $( param );

    if ( this.settings.onfocusout && target.not( ".validate-greaterThanEqual-blur" ).length ) {
        target.addClass( "validate-greaterThanEqual-blur" ).on( "blur.validate-greaterThanEqual", function() {
            $( element ).valid();
        } );
    }

	var val = target.val(), normalizer;
	if (rules = target.rules()) {
[...]

PR #2451, opened 2022-10-26
Added normalizations before comparing

enboig:
#### Description

It normalizes values before comparing them, and solves #2450


enboig:
Added a test. This could also be added to `equalTo` in `core.js`.

hackitnow278:
It normalizes values before comparing them, and solves#2450

hackitnow278:
> It normalizes values before comparing them, and solves#2450

PR #2453, opened 2022-11-06
 Localization: Add Hindi translation

saurabhsharma2u:
#### Description
 Localization: Update Hindi (India) translation

Thank you!


src/localization/messages_Hi_IN.js
```diff
@@ -0,0 +1,44 @@
+/*
+ * Translated default messages for the jQuery validation plugin.
+ * Locale: Hi-IN (Hindi; हिंदी)
```
bytestream, line 3:
```suggestion
 * Locale: Hindi; हिंदी
```

bytestream (changes requested):
I believe the file should be in ISO 639-1 format, so `message_hi.js`

src/localization/messages_Hi_IN.js
saurabhsharma2u: @bytestream Updated the suggested changes

bytestream (approved):

saurabhsharma2u:
@bytestream thank you for review, please let know if anything else is required to merge request

PR #2456, opened 2022-11-21
alphanumeric: fix the french error message; change regex pattern to accept french characters

polmih:
alphanumeric validation: 

- fix the french error message; 
- change regex pattern to accept french characters;

Thank you!


src/additional/alphanumeric.js
```diff
@@ -1,3 +1,3 @@
 $.validator.addMethod( "alphanumeric", function( value, element ) {
-	return this.optional( element ) || /^\w+$/i.test( value );
+	return this.optional( element ) || /^[a-zàâçéèêëîïôûùüÿæœ0-9_]+$/i.test( value );
```
bytestream, line 2: We probably want the JavaScript equivalent of `\p{L}\p{M}` - https://github.com/yury-dymov/js-regex-pl/blob/ff10757b2a98ad0bf0f62acebad085bab3748fc4/src/index.js

bytestream (changes requested):
Also needs a test

PR #2460, opened 2022-11-30
Removed === compare, changed to ==

wewhite:
Compare by value and type (===) does not work for this algorithm, as both cd and cdv can be either types at the same time. By comparing by value only (==) cd and cdv can be either integer or string, as a string number will be converted to a number reqardless of type.

Relates to BUG: https://github.com/jquery-validation/jquery-validation/issues/2459

### DIFF
--- a/src/additional/vinUS.js
+++ b/src/additional/vinUS.js
@@ -25,6 +25,7 @@ $.validator.addMethod( "vinUS", function( v ) {
 		f = FL[ i ];
 		d = v.slice( i, i + 1 );
 		if ( i === 8 ) {
+			// check digit vin is an integer
 			cdv = d;
 		}
 		if ( !isNaN( d ) ) {
@@ -35,6 +36,7 @@ $.validator.addMethod( "vinUS", function( v ) {
 					d = VL[ n ];
 					d *= f;
 					if ( isNaN( cdv ) && n === 8 ) {
+						// check digit vin is a string
 						cdv = LL[ n ];
 					}
 					break;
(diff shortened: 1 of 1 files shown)


src/additional/vinUS.js
```diff
@@ -43,11 +45,15 @@ $.validator.addMethod( "vinUS", function( v ) {
...
+    // check digit is an integer
+    cd = rs % 11;
 	if ( cd === 10 ) {
-		cd = "X";
+		// check digit is a string
+        cd = "X";
 	}
-	if ( cd === cdv ) {
+	// cd and cdv can be either integer  string
+	if ( cd == cdv ) {
```
bytestream, line 56: I think `parseInt()` should be used instead. For example, `(parseInt("X") === parseInt("X")) === false`. Any complaint against that?

```suggestion
	if ( parseInt(cd) === parseInt(cdv) ) {
```

bytestream (changes requested):
Please could you add a test, and revert the other changes as line 50 (`if ( parseInt(cd) === parseInt(cdv) ) {`) seems to be the only relevant change.

src/additional/vinUS.js
wewhite: No, that will not work, if string cd ="X" and string cdv="X" should = true.
line 50 could be 'if (+cd === +cdv')' the plus(+) on a string will evaluate it to an int, other wise it is a string.

In reality, the vin check digit in location 8 should be transliterated to a number not a string.
That way the check digit cd and check digit vin will always be a number.

wewhite:
I've rewritten the for loop, and removed the nested loop, different commit on my branch.
Will create a test.

wewhite:
Code is better now, removed nested for loop with an array index lookup.
Created test cases, added to the test/index file, and created a separate additional vinUS.js test file.

bytestream (approved):

bytestream:
Thanks :heart:

PR #2462, opened 2023-01-25
Core: Added escapeHtml option to avoid XSS attacks via showLabel method

volkanceylan:
This adds a new constructor option "escapeHtml".

There is a script injection risk if the messages set via $.validator.messages etc. are originating from a user localizable dictionary, like a translation screen as the showLabel function uses .html() to set the label content. Even in cases where that is not an issue, if the message itself contains a format placeholder like "{0} is not valid value for this field", as the value passed to {0} is a user input, it still provides an opportunity for script injection attacks albeit at a lower risk.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -945,14 +945,23 @@ $.extend( $.validator, {
 				error.removeClass( this.settings.validClass ).addClass( this.settings.errorClass );
 
 				// Replace message on existing label
-				error.html( message );
+				if ( this.settings && this.settings.escapeHtml ) {
+					error.text( message || "" );
+				} else {
+					error.html( message || "" );
+				}
 			} else {
 
 				// Create error element
 				error = $( "<" + this.settings.errorElement + ">" )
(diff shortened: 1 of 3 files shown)


bytestream:
Thanks @volkanceylan

bholmesACR:
@volkanceylan Hey there, did this fix get added to a new release? According to Sonatype vulnerability listing "sonatype-2023-0681", which links to this page, the vulnerability has not been addressed in a new release since the current one version 1.19.5 is still affected correct?

bytestream:
@bholmesACR see [1.20.0](https://github.com/jquery-validation/jquery-validation/releases/tag/1.20.0)

bholmesACR:
@bytestream Thanks! Also do we have to turn on/enable something like @volkanceylan says even after updating to remediate the risk? Just want to confirm

bytestream:
@bholmesACR You have to set `escapeHtml: true,` in your jquery-validation config

PR #2470, opened 2023-08-17
Localization: Rename message_hi.js to messages_hindi.js

EricDunsworth:
This makes it consistent with other i18n file names (which are all written in plural).


bytestream:
I think we'll have to hold off on that as it may break build systems that rely on the old file name.

EricDunsworth:
@bytestream Not sure if this makes much of a difference, but FWIW no official releases currently use the i18n file this PR is seeking to rename.

The [latest release of the jQuery validation plugin](https://github.com/jquery-validation/jquery-validation/releases) (1.9.5) was rolled out in July 2022 and doesn't include any Hindi translations at all. The ``src/localization/message_hi.js`` file was first introduced later on (November 2022 via #2453).

bytestream:
Thanks for the clarification @EricDunsworth

bytestream (approved):

PR #2472, opened 2023-08-17
Localization: Add Tagalog translations

EricDunsworth:
Notes:
* Only have a handful of translated messages on hand
* Improvised the max message based on min (renamed "malaki" to "maliit")
  * Should be comparable to renaming "greater" to "less"

### DIFF
--- a/src/localization/messages_tl.js
+++ b/src/localization/messages_tl.js
@@ -1,0 +1,11 @@
+/*
+ * Translated default messages for the jQuery validation plugin.
+ * Locale: TL (Tagalog; Tagalog)
+ */
+$.extend( $.validator.messages, {
+	required: "Ang field na ito ay required.",
+	email: "Mangyaring magpasok ng isang valid na email address.",
+	number: "Mangyaring magpasok ng isang valid number.",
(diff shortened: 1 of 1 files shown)


src/localization/messages_tl.js
```diff
@@ -0,0 +1,11 @@
+/*
+ * Translated default messages for the jQuery validation plugin.
+ * Locale: TL (Tagalog; Tagalog)
```
EricDunsworth, line 3: Should "Tagalog" be written only once at this spot (since AFAIK it's written the same way in both English and Tagalog)?

src/localization/messages_tl.js
```diff
@@ -0,0 +1,11 @@
...
+ */
+$.extend( $.validator.messages, {
+	required: "Ang field na ito ay required.",
+	email: "Mangyaring magpasok ng isang valid na email address.",
+	number: "Mangyaring magpasok ng isang valid number.",
+	max: $.validator.format( "Mangyaring magpasok ng value na mas maliit o katumbas ng {0}." ),
```
EricDunsworth, line 9: **Full disclosure:** I can't speak Tagalog (nor Filipino).

I improvised this translation because I felt like it wouldn't be ideal to provide the ``min`` translation I had on hand whilst omitting a ``max`` counterpart.

PR #2483, opened 2024-03-04
Core: Allow negative decimal with no 0

cstieg:
#### Description

Thank you!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1483,7 +1483,7 @@ $.extend( $.validator, {
 
 		// https://jqueryvalidation.org/number-method/
 		number: function( value, element ) {
-			return this.optional( element ) || /^(((?:-?\d+|-?\d{1,3}(?:,\d{3})+)|-\.\d+)?(?:\.\d+)?)$/.test( value );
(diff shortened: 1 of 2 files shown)


src/core.js
```diff
@@ -1483,7 +1483,7 @@ $.extend( $.validator, {
 
 		// https://jqueryvalidation.org/number-method/
 		number: function( value, element ) {
-			return this.optional( element ) || /^(?:-?\d+|-?\d{1,3}(?:,\d{3})+)?(?:\.\d+)?$/.test( value );
+			return this.optional( element ) || /^(((?:-?\d+|-?\d{1,3}(?:,\d{3})+)|-\.\d+)?(?:\.\d+)?)$/.test( value );
```
bytestream, line 1486:
```suggestion
			return this.optional( element ) || /^(?:-?\d+|-?\d{1,3}(?:,\d{3})+)?(?:-?\.\d+)?$.test( value );
```

Shouldn't it literally just be this?

bytestream (approved):
I suppose given `-123,000.00` and `.100` are valid, then `-.100` should be valid too.

PR #2486, opened 2024-04-08
Adding matriculaCertidaoBR and tituloEleitorBR validations

lucaslm:
#### Description

Adding new aditional validations for a few brazilian documents.
Thank you!


bytestream:
Thanks for the high quality PR! However, unless there is demand for these rules from others I think you're better maintaining these yourself otherwise it just extends the maintenance burden of the library.

PR #2493, opened 2024-06-26
Core: Added support for Web Components. Fixes #2492

### LINKED ISSUE #2492: Validate Web Components
## New feature motivation

With the rise of [Web Components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components) and the [ElementInternals API](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) it has become quite easy to build custom form elements. Currently, jquery-validation does not support Web Components which means the library does not validate but skip Web Components entirely when they are introduced into an already existing web application (our use case)

## New feature description

jquery-validation would need to allow custom tags to be validated as well

## New feature implementation

Currently, jquery-validation "only" allows the following elements to be considered in a form:

```
return $( this.currentForm )
    .find( "input, select, textarea, [contenteditable]" ) // <-- these selectors here
    .not( this.settings.ignore )
```
As I don't know a way to automatically include all "registered Web Components which also implements the ElementInternals API", the idea is to allow the user to define their own selector.

Similar to the element selector above, the selectors for event triggering would also need to be defined by the user:
```
$( this.currentForm )
    .on( "focusin.validate focusout.validate keyup.validate",
	   :text, [type='password'], [type='file'], select, textarea, ..... // <- here
         )
   .on( "click.validate", "select, option, [type='radio'], [type='checkbox']" // <- here
```
**Extending the validator options:**
We could either go with 3 new entries in the settings like:
```
$.extend( $.validator, {
	defaults: {
            ...
            ignoreTitle: false,
            listenFocusOn: ":text, [type='password'], ..., custom-text" // <-- new
            listenClickOn: "select, option, [type='radio'], [type='checkbox'], custom-text", // <-- new
            elements: "input, select, textarea, [contenteditable], custom-text", // <-- new
            onfocusin: function( element ) {
            ...
```
or we try to go the "smart way" and provide a single settings entry like:
```
$.extend( $.validator, {
	defaults: {
            ...
            ignoreTitle: false,
            customElements: [ "custom-text" ] // <-- new
            ...
```
I opened a PR with the single settings entry variant.

dhobi:
### Checklist for this pull request
- [x] Your code should contain tests relevant for the problem you are solving. 
- Test is included 
- [x] Your commits messages format should follow the jQuery git commit message format (http://contribute.jquery.org/commits-and-pull-requests/#commit-guidelines).
- Same as title
- [x] The pull request should reference existing issues or link to a reproducible demo.
- Referenced issue (feature request) is #2492

#### Description
This PR tries to add validation support for custom elements (Web Components). See #2492 for a first technical analysis.

It's possible that the current implementation is not good enough. Especially the part where I added the customElements to both, the "focusin.validate focusout.validate keyup.validate" events and to the "click.validate" events. Maybe we would need to try to differ there or we should go with "more settings value" mentioned in #2492 approach?

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,5 +1,8 @@
 {
 	"preset": "jquery",
 	"maximumLineLength": null,
-	"requireCamelCaseOrUpperCaseIdentifiers": null
+	"requireCamelCaseOrUpperCaseIdentifiers": null,
+	"excludeFiles": [
+		"test/custom-elements.js"
+	]
 }
--- a/.jshintignore
+++ b/.jshintignore
@@ -4,3 +4,4 @@ test/qunit/
 dist/
 demo/
 *.min.js
+test/custom-elements.js
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -115,7 +115,17 @@ grunt.initConfig( {
 		}
 	},
 	qunit: {
-		files: "test/index.html"
+		files: "test/index.html",
+		options: {
+			puppeteer: {
+				args: [
+					"--headless",
+					"--disable-web-security",
+					"--allow-file-access-from-files"
+				]
+			},
+			timeout: 10000
+		}
 	},
 	jshint: {
 		options: {
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
 		"grunt-contrib-concat": "1.0.1",
 		"grunt-contrib-copy": "1.0.0",
 		"grunt-contrib-jshint": "1.0.0",
-		"grunt-contrib-qunit": "1.2.0",
+		"grunt-contrib-qunit": "10.0.0",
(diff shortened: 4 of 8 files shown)


bytestream:
This looks OK to me.

> It's possible that the current implementation is not good enough.

Why? The way you've done it would be my preference over several settings.

dhobi:
> Why? The way you've done it would be my preference over several settings.

I was thinking because customElements are now listening to both event types: "focusin.validate focusout.validate keyup.validate" but also to "click.validate" instead of just either one of them. The only reliable way to "split" imo would be to make it configurable by the user (therefore the single vs multiple config entries dilemma). 
If we just say that web components will always listen for all events, that would be fine for me as well. It's just a decision in the end 😄

bytestream (approved):

PR #2494, opened 2024-06-29
Update cnhBR.js

rmff:
fix: checks for optional field (not required)

#### Description

Thank you!

### DIFF
--- a/src/additional/cnhBR.js
+++ b/src/additional/cnhBR.js
@@ -2,8 +2,13 @@
  * Brazillian CNH number (Carteira Nacional de Habilitacao) is the License Driver number.
  * CNH numbers have 11 digits in total: 9 numbers followed by 2 check numbers that are being used for validation.
  */
-$.validator.addMethod( "cnhBR", function( value ) {
-
+$.validator.addMethod( "cnhBR", function( value, element ) {
+  
+  // Field not required
+  if ( value.length === 0 && this.optional( element ) ) {
+	return true;
+  }
+	
   // Removing special characters from value
   value = value.replace( /([~!@#$%^&*()_+=`{}\[\]\-|\\:;'<>,.\/? ])+/g, "" );
 


bytestream:
This would change the way validation worked in previous versions, and it's subjective whether it should be optional or not. You'd be better off writing your own custom rule I think.

rmff:
Hi @bytestream,
I suggested this change because it's impossible to use cnhBR on an optional (not required) field, the validation always fails for empty values, even with `required: false`

I had use the same approach of other validations like:
https://github.com/jquery-validation/jquery-validation/blob/master/src/additional/cnpjBR.js
https://github.com/jquery-validation/jquery-validation/blob/master/src/additional/cpfBR.js
https://github.com/jquery-validation/jquery-validation/blob/master/src/additional/ipv4.js
https://github.com/jquery-validation/jquery-validation/blob/master/src/additional/bankaccountNL.js
https://github.com/jquery-validation/jquery-validation/blob/master/src/additional/cifES.js
etc...

Most part of additional validations uses 
`
	if ( this.optional( element ) ) {
		return true;
	}
`
for check if the filed is optional.

Getting deeper I understand that add  `value.length === 0` maybe redundant and seems to be already treated by `this.optional( element )`

Can I make an new pull request fixing it, aligned with the most part of additional validations? 

I'll update to include
`
	if ( this.optional( element ) ) {
		return true;
	}
`
on the beginning of the method.

lucaslm:
Hey there, @rmff.

I actually had the same problem with another rule (ie, `nisBR`. It turns out it also fails when the field is empty.

However, I could still use it the way I intended it, allowing empty values but still validating everything else.

If you look at the docs [here](https://jqueryvalidation.org/validate/), you will notice there is a way to specify a rule as an plain object, having a _`"depends-property to apply the rule only in certain conditions"`_ (in our cases, not empty values).

An exemple would go along the lines of:

```
'yourElementName': {
  cnhBR: {
    depends: function(element) {
      var value = $(element).val() || '';
      return value.trim().length;
    }
  },
  //other rules, such as required, etc
},
```

Neat, right?

PR #2498, opened 2024-09-04
Core: Validate inputs with duplicate name / name[] arrays, instead of just the first item

bompus:
#### Description

Replaces #2000 and #2431 with simpler logic, and I believe that .elements() should return all of the eligible elements.

Previously, if you had 2+ inputs using name arrays, such as:
```html
<input type="text" name="todo_item[]" data-rule-required="true" />
<input type="text" name="todo_item[]" data-rule-required="true" />
<input type="text" name="todo_item[]" data-rule-required="true" />
```

It would only hit validation on the first input in the array, allowing all the remaining inputs with the same name to remain empty.

Technically, I'm pretty sure the following, without using [] syntax would have the same problem:
```html
<input type="text" name="first_name" data-rule-required="true" />
<input type="text" name="first_name" data-rule-required="true" />
<input type="text" name="first_name" data-rule-required="true" />
```

I'm not familiar with the testing framework, but please review the idea behind this PR first, and if you're confident that it will be merged if tests are included, I'll dig deeper into testing.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -453,9 +453,22 @@ $.extend( $.validator, {
 		},
 
 		checkForm: function() {
+			var checkedCache = {};
+			
 			this.prepareForm();
 			for ( var i = 0, elements = ( this.currentElements = this.elements() ); elements[ i ]; i++ ) {
-				this.check( elements[ i ] );
+				var element = elements[ i ];
+				
+				var type = element.attr( "type" );
+				if ( type === 'checkbox' || type === 'radio' ) {
+					// only check the first checkbox/radio input ( https://github.com/jquery-validation/jquery-validation/pull/2431#issuecomment-1172835268 )
+					if ( element.name in checkedCache ) {
+						continue;
+					}
+					checkedCache[ element.name ] = true;
+				}
+				
+				this.check( element );
 			}
 			return this.valid();
 		},
@@ -657,8 +670,12 @@ $.extend( $.validator, {
 					return false;
 				}
 
-				// Select only the first element for each name, and only those with rules specified
-				if ( name in rulesCache || !validator.objectLength( $( this ).rules() ) ) {
+				if ( name in rulesCache ) {
+					return true;
+				}
+
+				// return only those with rules specified
+				if ( !validator.objectLength( $( this ).rules() ) ) {
 					return false;
 				}
 
--- a/src/core.js
+++ b/src/core.js
@@ -457,7 +457,7 @@ $.extend( $.validator, {
 			
 			this.prepareForm();
 			for ( var i = 0, elements = ( this.currentElements = this.elements() ); elements[ i ]; i++ ) {
-				var element = elements[ i ];
+				var element = elements.eq(i);
 				
 				var type = element.attr( "type" );
 				if ( type === 'checkbox' || type === 'radio' ) {


src/core.js
```diff
@@ -453,9 +453,22 @@ $.extend( $.validator, {
...
+				var element = elements.eq(i);
+				
+				var type = element.attr( "type" );
+				if ( type === 'checkbox' || type === 'radio' ) {
+					// only check the first checkbox/radio input ( https://github.com/jquery-validation/jquery-validation/pull/2431#issuecomment-1172835268 )
+					if ( element.name in checkedCache ) {
+						continue;
+					}
+					checkedCache[ element.name ] = true;
+				}
```
KES777, line 469: This is not clear why we check only first checkbox?

What about the cases when we ask user to check boxes which labels does not have `a` letter?

src/core.js
```diff
@@ -657,8 +670,12 @@ $.extend( $.validator, {
 					return false;
 				}
 
-				// Select only the first element for each name, and only those with rules specified
-				if ( name in rulesCache || !validator.objectLength( $( this ).rules() ) ) {
+				if ( name in rulesCache ) {
+					return true;
+				}
```
KES777, line 675: Actually this is correct, that we do not allow elements with the same name. I believe this is by RFC. Though I could be wrong, because I did not work with checkbox/radio a lot.

KES777:
@bytestream  I am not familiar with checkbox/radio input types, but it looks like they try to resolve the different problem.

src/core.js
bompus: Checkbox/radio inputs that share a `name` are one control (one submitted value / one required group). The plugin already validates that group through the first input (`validationTargetFor` / `findByName`). Running `check()` on every box in the group would apply the same rule twice.

Per-checkbox rules that depend on the label (for example “must not contain the letter a”) need distinct `name`s. That is a different feature from this PR.

This first-per-name group behavior is the same one called out in https://github.com/jquery-validation/jquery-validation/pull/2431#issuecomment-1172835268. #2512 (globbing `Company[Phone][n][phone]` onto one rule) is also a different problem.

src/core.js
bompus: Same-name text inputs (especially `name[]` arrays) are valid HTML and a common PHP/Rails pattern. The old `elements()` filter kept only the first field with that name, so later required inputs were never checked.

This PR keeps every same-name text/select/textarea that has rules, and still treats checkbox/radio as one group (first input only).

bompus:
@bytestream @KES777 Rebased onto current `master` and added the tests I deferred in the original description.

What changed:
- `elements()` now returns every same-name text/select/textarea that has rules (`name[]` and duplicate names), so later required fields are actually validated.
- Checkbox/radio groups stay first-only (existing group behavior; see https://github.com/jquery-validation/jquery-validation/pull/2431#issuecomment-1172835268). `checkForm` caches by the DOM `name`, so two groups are not collapsed into one.
- QUnit coverage for `name[]`, duplicate text names, and two checkbox/radio groups. `grunt` jscs/jshint/qunit is green (295 tests).

This is not #2512 (rule globbing for `Company[Phone][n][phone]`). Happy to adjust if you want a different split between group controls and array fields.

KES777:
Good news, thank you @bompus for your work. Unfortunately I have not worked with this plugin recently, thus I can not test the changes.

PR #2504, opened 2024-10-14
Allow AVIF and WebP files

vanillajonathan:
Allow files with the extensions `.avif` and `.webp`.


bytestream:
Can you not just use the `accept` extension?

vanillajonathan:
I suppose I could but this seems like a reasonable PR. AVIF and WebP are now common on the web so where PNG, JPEG and GIF are accepted AVIF and WebP are expected to be accepted too.

bytestream:
Someone can correct me if I'm wrong, but after reviewing the Git history I don't think there is precedent for updating existing additional rules. IMO these changes constitute a breaking change and would therefore need to wait until v2. I think you're better off just doing what I suggested, or define your own rule...

PR #2517, opened 2025-06-04
Localization: Additional string translations for messages_nl.js

appel:
#### Description
I don't know if you accept pull requests, but I noticed messages_de.js had a ton of strings that were missing from messages_nl.js, so I've added translations for those additional strings to messages_nl.js.


appel:
@bytestream Apologies, should be fixed now.

PR #2535, opened 2026-04-01
Security: mitigate CVE-2015-9251 in vendored jQuery (gh-2432)

### LINKED ISSUE #2432: Incorrect date of 1.19.5 release
Hello.

I was waiting for this **1.19.5** release. Got a notification that it was released today. Yet it seems that release date of version should actually be **2022-07-01**, but in [Releases](https://github.com/jquery-validation/jquery-validation/releases) and [changelog.md](https://github.com/jquery-validation/jquery-validation/blob/master/changelog.md) I find **2022-06-01**.

Is this intentional?

navnitan-7:
## Summary

Mitigate **CVE-2015-9251** (jQuery gh-2432) in the vendored jQuery copies under **`lib/`** (`jquery-1.11.1.js`, `jquery-3.1.1.js`, `jquery-4.0.0.js`) by adding the same `ajaxConvert` guard as upstream: skip cross-domain auto-`script` conversion unless explicitly intended.

## Upstream reference

- [jquery/jquery@2546bb35](https://github.com/jquery/jquery/commit/2546bb35b89413da5198d54a4539e4ed0aaf6e49) — *Ajax: Mitigate possible XSS vulnerability* (Fixes gh-2432)

## Changes

- Insert `if ( s.crossDomain && current === "script" ) { continue; }` in `ajaxConvert` for each listed file (before seeking converters).
- Add **`scripts/verify-jquery-cve-2015-9251.sh`** so CI or local runs can assert the guard string is present in all three copies.

## Verification

```bash
bash scripts/verify-jquery-cve-2015-9251.sh
npx grunt
```

- Verify script: **ok** (expects guard in all three `lib/jquery-*.js` files).
- **Grunt:** JSCS, JSHint, QUnit — **290 tests, 0 failed** (local run).

### DIFF
--- a/lib/jquery-1.11.1.js
+++ b/lib/jquery-1.11.1.js
@@ -8767,6 +8767,11 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
 			// Convert response if prev dataType is non-auto and differs from current
 			} else if ( prev !== "*" && prev !== current ) {
 
+				// Mitigate possible XSS vulnerability (gh-2432)
+				if ( s.crossDomain && current === "script" ) {
+					continue;
+				}
+
 				// Seek a direct converter
 				conv = converters[ prev + " " + current ] || converters[ "* " + current ];
 
--- a/lib/jquery-3.1.1.js
+++ b/lib/jquery-3.1.1.js
@@ -8682,6 +8682,11 @@ function ajaxConvert( s, response, jqXHR, isSuccess ) {
 			// Convert response if prev dataType is non-auto and differs from current
 			} else if ( prev !== "*" && prev !== current ) {
 
+				// Mitigate possible XSS vulnerability (gh-2432)
+				if ( s.crossDomain && current === "script" ) {
+					continue;
+				}
+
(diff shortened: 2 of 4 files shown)


bytestream:
These files are only used for testing so I'm not sure this is really necessary. Best case, we remove the files and include them via package manager but that might be a bit involved.

navnitan-7:
Thank you for the clarification.

You're right that these files are only used for testing. My intention with this change was to keep dependencies up to date and aligned with current versions, even in test environments, to avoid potential security or maintenance issues down the line.

That said, I agree this may not be strictly necessary if the files are isolated to tests. I'm also open to the alternative approach of removing the vendored files and managing them via a package manager if that better fits the project's direction.

Please let me know your preference, and I'm happy to adjust the PR accordingly.

bytestream:
>  I'm also open to the alternative approach of removing the vendored files and managing them via a package manager if that better fits the project's direction.

If you'd like to have a go at that, please do 👍🏼  I think we can support jQuery 3 and 4 given they're the only supported versions.

